PR #7, opened 2018-07-08
Moved error message for missing auth token

### LINKED ISSUE #4: packet-cli help in the absence of a token or config file
The help command requires an authentication token or config file to run, which is a challenge if you're trying to get help about this particular requirement.

```
Eds-MacBook-Pro:bin ed$ ./packet-cli help
Packet authentication token not provided. Please either set the 'PACKET_TOKEN' environment variable or create a JSON configuration file.
```

I think this actually may be the correct behavior to insist on getting a token or config file, but the message returned might also return the normal help text in addition, since generating that text doesn't require the token.

jasmingacic:
Closes #4 

p.s. Test binaries are available in the attached zip archive

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -95,7 +95,7 @@
   branch = "master"
   name = "github.com/packethost/packngo"
   packages = ["."]
-  revision = "db5b2fb99ee303eba5699cb0f9d74aeba8b39b73"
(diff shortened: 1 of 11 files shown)


vielmetti (commented):
Looks good to me.

PR #13, opened 2019-01-10
Update Makefile to 0.0.2

gregkh:
VERSION needs to be bumped to 0.0.2 in the main makefile.


t0mk:
@gregkh Why the version needs to be bumped?

gregkh:
Because you bumped the version in a much older patch in other places, why not here as well for when you want to make a released image?

jasmingacic:
Latest released version is 0.0.2 and the makefile does not reflect that, also there is no harm in doing this.

PR #17, opened 2019-07-12
fix: non-json config in homedir

### LINKED ISSUE #14: yaml config file not autodetected
Following the readme suggests one could use either yaml or json config file and it should be detected by default. This is not the case (only the JSON file will be detected).

I'm happy to PR to autodetect the yaml file in the indicated location (`$HOME/.packet-cli.yaml`) if that would be accepted?

alexeldeib:
Fixes https://github.com/packethost/packet-cli/issues/14

Adds err check on file read for config initialization.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -22,7 +22,6 @@ package cmd
 
 import (
 	"fmt"
-	"io"
 	"os"
 	"runtime"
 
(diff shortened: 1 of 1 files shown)


t0mk:
thanks for the PR @alexeldeib, looksgood!

jasmingacic:
This fix is breaking the CLI each command that is ran without a config file results in this:
```
packet plan get 
Unable to read in config file, check the file path and defaults (./packet-cli.{yaml,json} in user home directory) and try again.
```
It seems even when config file is not present it overrides the environment variable.

jasmingacic:
Also the message is incorrect. The file created should be .packet-cli.{yaml,json} not ./packet-cli.{yaml,json} 
This does not reflect what users are supposed to do according to the docs `https://github.com/packethost/packet-cli#authentication`

PR #19, opened 2019-07-12
feat: allow retrieving projects by name

alexeldeib:
Partially addresses https://github.com/packethost/packet-cli/issues/15

### DIFF
--- a/cmd/create_device.go
+++ b/cmd/create_device.go
@@ -29,6 +29,7 @@ import (
 )
 
 var (
+	projectName     string
 	projectID       string
 	facility        string
 	plan            string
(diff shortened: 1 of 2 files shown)


t0mk:
Hi, @alexeldeib thanks for the PR. I took a look.

Can you makre the CLI err if the given name doesn't match any existing project?

It errs on wrongs UUID (404 fro the API), it would be nice if the behavior would be the same for the lookup by name.

alexeldeib:
@t0mk do you think it's better like this as a separate flag, or try parse as uuid and fall back to name, e.g.:

```go
_, err := uuid.FromString("not-a-uuid")
if err != nil {
  // Try fetch using name, maybe still 404
}
// Try with ID
```
might be better to try parse and avoid using flag namespace

t0mk:
@alexeldeib no, that might be a bit confusing. Let's keep in in separate flags. It looks good, I am merging.

PR #21, opened 2019-08-05
fix: allow failure reading config file

### LINKED ISSUE #20: Previous PR breaks the CLI
Changes submitted [here](https://github.com/packethost/packet-cli/pull/17) are breaking the CLI. 
It does not allow users to use only environment variables, but only config files. This does not comply with the instructions https://github.com/packethost/packet-cli#authentication

alexeldeib:
Fixes https://github.com/packethost/packet-cli/issues/20

Sorry! 😢 @jasmingacic

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -86,7 +86,7 @@ func initConfig() {
 		viper.AddConfigPath(userHomeDir())
 	}
 
-	_ = viper.ReadInConfig()
+	viper.ReadInConfig()
 
 	if viper.GetString("token") != "" {
 		packetToken = viper.GetString("token")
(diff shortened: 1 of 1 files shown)


jasmingacic:
@alexeldeib The fix seems to be ok. 
But the question is why do you need this implementation when the CLI was already accepting json and yaml config files?

alexeldeib:
It wasn't accepting yaml files. That was the exact point of the fix.

jasmingacic:
Yeah that makes sense. 
I would only replace this:
```
_ = viper.ReadInConfig()
```
with this
```
viper.ReadInConfig()
```

alexeldeib:
I'm happy either way, it's uglier but I tend to prefer explicitly vs implicitly ignoring the error. Will let @t0mk decide preference since he's stuck maintaining it 😃

t0mk:
Yes let's drop the undescore. I will do it.

If eveything is OK, shall we merge?

jasmingacic (approved):

jasmingacic:
Go ahead get it merged.

PR #23, opened 2019-08-28
Fixing the order of 'get' and subcommand

nqn:
Simple nit: the documentation and command line help needs the order of subcommands changed to match the current CLI.


jasmingacic (approved):

jasmingacic:
Thanks, good catch.

Issue #34, opened 2019-11-24
Rename "device" to "server"?


rusenask:
I think it's a bit unusual to call a server "device", in the UI it looks like:

![image](https://user-images.githubusercontent.com/9080105/69502497-f3576480-0f07-11ea-9516-16c6d7e3dcb6.png)

When I first got the CLI I just couldn't see any command that would help me list the servers.

PR #36, opened 2019-11-25
Various fixes

### LINKED ISSUE #27: Go mod support
Would you accept a PR to move to using Go modules? It seems the community is shifting in that direction, and I must say, for all the hubbub, it has been a pleasant experience to use compared to other Go tools I'd used previously.

### LINKED ISSUE #31: --termination-time rejects values that the API accepts
As I'm trying to adopt `packet-cli` over my pile of shell scripts, I've noticed that the termination_time I was calculating and passing directly to the API is rejected by the CLI. This seems unfortunate, to me.

It would be nice if the same values could be passed to the CLI. Better yet, if you could pass things like `--termination-time="+2 hours"` would be very cool as well, since this is what I'm doing anyway:

```bash
# should work, but doesn't:
termtime="$(TZ=UTC date --date="2 hour" --iso-8601=seconds)"
packet device create --termination-time="${termtime}" ...

# maybe in the future?
packet device create --termination-time="+2 hours"
```

### LINKED ISSUE #32: `packet plan get` doesn't list any pricing info
IIRC the api returns pricing info about the various plans available, but this isn't included in the table or json/yaml output from the `packet plan get` command.

jasmingacic:
GO modules support (closes #27)
Fixes #31 
Fixes #32


jasmingacic:
Fixed #34

PR #42, opened 2019-12-18
build FreeBSD as well

dch:
- builds on FreeBSD 12.1R amd64 & 13.0-CURRENT amd64
- will check arm64 / armv8 later
- please confirm if this cross-builds in your CI or not


jasmingacic:
I just got this error
```
cmd/go: unsupported GOOS/GOARCH pair freebsd/arm64
```

Other than that it seems to be working

jasmingacic (approved):

dch:
Thanks looks like we need to wait a couple months for go 1.14 to land and then we are good. https://github.com/golang/go/issues/24715

PR #60, opened 2020-07-01
Adds shell completions and custom user-agent

### LINKED ISSUE #55: Packet CLI does not set an identifying User-Agent
## What problem are you facing?

As users of Packet tools, API usage by these tools should be visible to Packet.  We understand that this kind of information can lead to recognition and better support for specific tools.  Support could come in the form of developer time on front-end or back-end aspects of the tools, addressing bugs, adding features, or improving performance.

## How can packet-cli help?

All API consumers should be instrumented with an appropriate [User-Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent). The UA is included in HTTP calls to the Packet API.  The precise User-Agent used in each tool will differ to some degree. The expected components of a UA should include the name and version of the tool itself as well as the primary plugins and libraries contributing to the tool's operation.

For example, 
https://github.com/packethost/packet-cli/blob/master/cmd/cli.go#L29 could benefit from a `cli.Client.UserAgent` assignment.  In this case, the name and version of the CLI and the name and version of `packngo` should be included.

```
User-Agent: packet-cli/0.0.8 packngo/0.0.2
```

This may benefit from packngo exposing the version to consumers or exposing functions like `SetUserAgent` or `PrependUserAgent` that enhances the visibility of related tools.

### LINKED ISSUE #56: Include Bash autocompletion scripts
## What problem are you facing?

As a user, I find auto-completion scripts help me to identify the available commands and parameters to commands.  I can use tab-completion to avoid breaking out of the command I am typing to uncover simple CLI usage details, like the name, spelling, hyphenation, or existence of a particular parameter.

The Packet CLI does not include auto-completion scripts for my shell, Bash.

## How can Packet help?

The CLI should include a command to generate the autocomplete scripts.  This can be implemented [here](https://github.com/packethost/packet-cli/blob/master/main.go#L35), taking advantage of spf-13/cobra support for the feature, described here: https://github.com/spf13/cobra/blob/master/shell_completions.md (more here https://github.com/spf13/cobra/blob/master/bash_completions.md).

Once the `packet-cli completion bash` (equivalent) command exists, [`make`](https://github.com/packethost/packet-cli/blob/master/Makefile#L36-L38
) should incorporate this.

The current approach to triggering generators is to set specific environment variables, this experience could be improved as command-line arguments.

### LINKED ISSUE #63: Commands are inconsistent about plural or singular names
It is difficult to build muscle memory and intuition about the Packet CLI commands become some commands are plural (facilities, operating-systems, payment-methods) while others are singular (plan, project, device).

SPF13 supports aliases.  Aliases should be added to make plural or singular alternatives for each command.

Aliases can also be used to provide short names for commands like "virtual-network" (vlan), "operating-systems" (os), and perhaps others.

displague:
* Adds shell completions `source <(packet completion bash)` / `eval "$(packet completion bash)"`
* Adds plural and singular command aliases for each command (this is inconsistent between resource types, adding the alternates resolves this)
* Moves doc generation to a command `packet docs ./docs` (instead of requiring an environment variable to be set)
* Includes packet-cli version in the User-Agent of the API requests

Closes #55 
Closes #56
Closes #63

---

```console
$ source <(packet completion bash)
$ # in Bash 3.2.x, you may need to use:  eval "$(packet completion bash)"
$ packet <tab>
2fa                   ip                    user
capacity              operating-systems     virtual-network
device                organization          volume
event                 plan                  vpn
facilities            project

$ packet device <tab>
create  delete  get     reboot  start   stop    update

$ packet device get --<tab>
--config       --config=      --id           --id=          --json
--project-id   --project-id=  --yaml
```


jasmingacic (approved):

jasmingacic:
@displague I will merge this

Issue #33, opened 2019-11-21
Variable scoping is a bit concerning


colemickens:
The codebase surprises me in some ways. For example, I was trying to understand how userdata was handled in `packet device create`. I was surprised to find that things like `alwaysPXE` or `hardwareReservationID` are effectively globals in the entire code base. I guess it doesn't wind up being an issue if the code is ever only used in this CLI context where it's always just one off invocations, but it still seems like a bit of a code smell. Has this been an issue in practice? What happens when two commands share flag names?

An example, when you look at "create_volume" you have to just know that most of the variables that are populated from the CLI  are defined in ... `create_device.go`. It would almost make more sense if all command flags were declared in a single file; and it would likely make it stick out more as a weird pattern, at least it does to me.

colemickens:
I'm more accustomed to seeing a struct paired with each command, and the command's args are parsed into that struct. That way the model and command are in the same source code file and easy to grok, and you don't have to worry about the fact that any code anywhere can touch "projectId" and affect any command that might be in flight. (For example, I've previously been able to create composite cobra commands that effectively invoke other commands internally in sequence or parallel -- that would be impossible to do with the current lack of isolation between commands/invocations).

jasmingacic:
We can refactor it. What are you suggesting?

rusenask:
It would be better to group command options into structs like and drop `init()` functions:

```golang
type createOptions struct {
	name          string
        // more fields..
}

// then don't use init() in the file, there's no point, you can get the flags
// in the function itself:
func newCreateCommand(cli *command.CLI) *cobra.Command {
	opt := createOptions{}

	cmd := &cobra.Command{
		Use:   "create [OPTIONS]",
		Short: "Create something",
		RunE: func(cmd *cobra.Command, args []string) error {
			return runCreate(cli, opt)
		},
	}

	flags := cmd.Flags()
	flags.StringVar(&opt.name, "name", "", "Runner name")
	
	return cmd
}

```

Then each command could have their own options and not reuse them in an unknown ways. Also, in the example above I also pass `command.CLI` struct that holds:

```golang
type CLI struct {
	client api.APIClient
	config *config.Config
}
```

so we could get rid of global API client for packet and config that can include things like formatting (would be nice to have an option to remove this table wrapper, set default project ID).

It would be easier to do this in steps, maybe just start moving command groups into their own directories so everything is not in one package?

jasmingacic:
[...]

PR #66, opened 2020-07-22
[WIP] refactor volumes to demonstrate how globals could be avoided

displague:
This experimental refactor of the `volumes` commands intends to demonstrate that global variables can be avoided with `spf13` usage.  

Testing these commands should be simpler because the argument and client variables are not set in a global context.

API mocks have been but these are not currently being used by tests (something that I would want to add to validate this approach).

Some of the commands have lost their `json` and `yaml` flags in this initial state.  I'm interested in seeing if these arguments can become global or if they are best suited for per-command-action use.

Relates to #33

---
Observations and Ponderances:

* this PR does not break the cli, only volumes are refactored and the rest of the project is mostly untouched
* `cmd/*volume*.go` moves to a `volume` package in `internal/volumes/`.
* When everything else has moved to `internal/`, `cmd/packet` can be created, which would facilitate https://github.com/packethost/packet-cli/issues/53. `go get` would then create binaries with the desired `packet` name, rather than `packet-cli`.
* Each command is explicitly registered, similar to what was is done currently in `init()` calls.
* `output` package and functions are created as a way to preserve the existing output functions while moving commands to their own package.
* ~Stderr is not used for error messages, neither is spf13's RunE~ #70 #75 
* license headers and go function and constant descriptions have not been added.
* I took advantage of spf13's `Example` field, rather than using the `Long` field. I can see why `Long` was used to portray examples, and I think this was to avoid the short description from being repeated in the help and generated markdown.  I think a better experience can be had by providing meaningful content in the long description fields.


displague:
All of the features presented in this draft have been incorporated, most notably by #133.  Closing.

PR #73, opened 2020-08-11
[WIP] add a Terraform module interface

displague:
This experimental branch adds a terraform module interface to `packet-cli`.
The goal of this is to make it easier to consume [Packet provider modules](https://registry.terraform.io/search/modules?q=equinix%2Fmetal) from the official Terraform registry.

Only github.com/packethost and github.com/packet-labs modules would be considered.

_Update: Only the Equinix Metal provider and Equinix modules would be considered)

## Current blockers

* [ ] Hashicorp/terraform does not expose a "modules list" function.

### Considerations

* [ ] This adds a ton of dependency. Can `packet terraform` be built as an independent plugin through a plugin interface (like git plugins)? (possibly a shell script?)

## Future workflow

Examples of use:

### list all Packet modules

```
$ packet terraform get
```

### show module details

The registry API includes the parameters of the module.

```
$ packet terraform get kubernetes-bgp
+----------------+---------+
|     SOURCE     | VERSION |
+----------------+---------+
| kubernetes-bgp | 0.0.1   |
+----------------+---------+
```

### create an instance of that module

These module parameters could be exposed as arguments.

This would effectively perform `terraform init --from-module=packet/k8s/packet`.
What directory would this go in? Should the base directory be common - and pass `workspace` through? (Likening this to `go get`, this would be like `go get registry/module@workspace`)

The api token would be wired-up, since the packet cli already knows one. How does it do this? Well known APIKey / Token variable names?

```
$ packet terraform create k8s --workspace foo --facility ewr1
```


displague:
> Hashicorp/terraform does not expose a "modules list" function.

Depite the limitations from https://github.com/hashicorp/terraform/tree/main/internal/initwd not being public, there are several workarounds:

* don't list modules, take whatever parameters are supplied and assume that equinix/$1/metal is a valid module.  take all `--` arguments that are not recognized and produce a terraform.tfvars with these values
* bundle the `terraform` plugin with known module help scripts, each can validate the parameters to the variables supported by the module. Each script can also translate packet-cli variables (token) to the name of the variable in each specific module (metal_auth_token, auth_token, etc).

displague:
@ctreatma found https://github.com/hashicorp/terraform-config-inspect which may be a suitable workaround for the https://github.com/hashicorp/terraform/tree/main/internal/initwd concerns mentioned [above](https://github.com/equinix/metal-cli/pull/73#issuecomment-809691417).

PR #74, opened 2020-08-12
Do not require token for completions subcommand

detiber:
Removes the requirement of having a packet api token configured
to run the completions subcommand


displague (approved):
Thanks, @detiber! 

I ran this without a token to see that `packet completion bash` did not error and that `packet devices get` returned an appropriate error.

Unrelated to your PR, I'll note that

PR #88, opened 2020-09-28
add device create --userdata-file argument

displague:
Quoting multi-line files for `packet device create --userdata` is common and not practical. Quotes are hard to maintain with complex scripts. Userdata can be supplied as a multipart mime or gzip content, which is easier to handle with a file argument than it is with a string argument.

This PR enables: `packet device create --userdata-file userdata.txt`

### DIFF
--- a/cmd/create_device.go
+++ b/cmd/create_device.go
@@ -22,6 +22,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
 	"time"
 
 	"github.com/packethost/packngo"
@@ -40,6 +41,7 @@ var (
 
 	storage               string
 	userdata              string
+	userdataFile          string
 	customdata            string
 	tags                  []string
 	ipxescripturl         string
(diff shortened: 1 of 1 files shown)


cmd/create_device.go
```diff
@@ -60,58 +62,46 @@ packet device create --hostname [hostname] --plan [plan] --facility [facility_co
...
-		}
-
-		if spotPriceMax != 0 {
-			request.SpotPriceMax = spotPriceMax
+		if userdataFile != "" {
+			userdataRaw, readErr := ioutil.ReadFile(userdataFile)
+			if readErr != nil {
+				return errors.Wrap(readErr, "Could not read userdata-file")
+			}
+			userdata = string(userdataRaw)
```
detiber, line 76: Might be good to add a check to see if the user specified both --userdata-file and --userdata to avoid any surprises if they specify both and only userdata-file is applied

displague: https://github.com/packethost/packet-cli/pull/88/files#diff-4dea4c1acc6ef47a6f4720eaa071fe90R67-R69

detiber: :grimacing: apologies, multitasking got the better of me.

detiber (approved):

PR #105, opened 2021-01-19
Added 'list' alias to get commands 

### LINKED ISSUE #101: Rename the CLI tool (and repo)
Packet is Equinix Metal.

This tool's name should be updated to match.

Here are some alternatives for the CLI command name:
* `metal`
* `equinix metal ...`
* `emctl`

jasmingacic:
Added `list` alias to `get` where the command name is in plural.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ K := $(foreach exec,$(EXECUTABLES),\
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
 PACKAGE_NAME?=github.com/packethost/packet-cli
-BINARY?=equinix
+BINARY?=packet
 GIT_VERSION?=$(shell git log -1 --format="%h")
 VERSION?=$(GIT_VERSION)
 RELEASE_TAG?=$(shell git tag --points-at HEAD)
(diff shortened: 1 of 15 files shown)


displague:
@jasmingacic Let's keep the get/list aliasing separate from the binary repo. Each of these changes will require additional changes:
* the binary rename requires documentation updates and the rename/movement of the project. I'm not sure about the desired name. I wouldn't want to take the `equinix` name for a tool that only functions on `equinix metal`. I'm more preferential to `metal` or `emctl` but would rather open this issue up to more folks before acting.
* `list` aliases will need to be updated in the documentation. I believe `make` will update the docs accordingly. What's in this PR seems right, otherwise.

cmd/retrieve_user.go
```diff
@@ -33,6 +33,7 @@ var (
 // retriveUserCmd represents the retriveUser command
 var retriveUserCmd = &cobra.Command{
 	Use:   "get",
+	Aliases: []string{"list"},
```
displague: This is a singular resource endpoint, I don't think `list` should be aliased here.

Makefile
```diff
@@ -5,7 +5,7 @@ K := $(foreach exec,$(EXECUTABLES),\
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
 PACKAGE_NAME?=github.com/packethost/packet-cli
-BINARY?=packet
+BINARY?=equinix
```
displague: Let's leave this to another PR.

jasmingacic:
Ok I removed binary rename. And I will rename the PR to something else.

Issue #54, opened 2020-06-15
Offer a way to set defaults using the config file


displague:
## What problem are you facing?

I often reuse the same settings when creating new resources using the Packet CLI.  I must provide these settings on each invocation of `packet-cli`.

## How can packet-cli help?

Common settings should be configurable per command, while high-level, commonly reused parameters, can be set globally. For example, in `~/.packet-cli.json`:

```json
{
  "token": "foo",
  "project-id": "some-project",
  "organization-id": "some-organization",
  "device": {
    "plan": "some-default-plan",
    "facility": "some-default-facility",
    "operating-system": "some-os",
  }
}
```

These parameters would coordinate with those made available through the CLI.

```
packet device create --hostname [hostname] --plan [plan] --facility [facility_code] --operating-system [operating_system] --project-id [project_UUID]
```

Parameters like `facility` may need more careful handling.  Should they be global or command-specific?  Should all parameters trickle down from global to command specific structures (Global facility is an option, but device-specific facility takes priority)?

Any form of default setting must be careful not to quash a user's ability to *not specify* an argument (default project-id should not affect a command that behaves differently if a project-id was not provided).

It would also be helpful to show the default values when displaying the help text, if possible.

displague:
https://github.com/packethost/packngo/issues/189 would offer these features generically through the Go client library, benefitting packet-cli, terraform, docker-machine-driver, and other tools.

displague:
It doesn't seem necessary to me to support the `~/.packet-cli.json` / `~/.packet-cli.yaml` file name.

displague:
I'm not convinced that we need to continue supportting both `yaml` and `json`.

displague:
defaults can now be provided in ~/.config/equinix/metal.yaml

We may want to restrict which fields can be used as defaults in the future (`id` should never be pulled from config).

PR #124, opened 2021-05-26
Updates root, env, relevant cmds with help text, to use rebrand langu…

jmarhee:
…age; Updates Makefile to change binary name to metal, updates reference to binary name to use that variable.

Signed-off-by: Joseph D. Marhee <jmarhee@equinix.com>

I will also try to address the config file changes in this PR as well, now that I've combed through the repo a bit; I think this should cover any user-facing instances of `packet` to `metal` (either in the binary name, or in help text, etc.) 

I built this without issue, and found the language updated as expected, so it will appear that way in the completed release.

Re: #101


jmarhee:
To clarify, I'm looking into the config file flow (both, format, and awareness of the name, etc.) and will add it in subsequent commits (hence the draft, since both issues are relatively straightforward), after addressing the source of the failing ci.

displague:
@jmarhee looks like you'll need `make clean-docs generate-docs; git add docs` (This is at least part of the failing CI.)

Makefile
```diff
@@ -5,7 +5,7 @@ K := $(foreach exec,$(EXECUTABLES),\
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
 PACKAGE_NAME?=github.com/packethost/packet-cli
```
displague: Perhaps we should replace this with github.com/equinix/metal-cli in a separate commit (and capture that change in go.mod too).

displague: It looks like we can replace `packethost/packet-cli` with `equinix/metal-cli` everywhere.
And then another replace to get `packet-cli` converted to `metal-cli` will pick up some stragglers.

jmarhee: Agreed. I wasn't sure if this was appropriate at this stage, but if it won't be a problem, then will do.

jmarhee:
Noting so I remember when I pick this up in the morning: My plan currently is, also, to have the CLI prefer `.metal-cli` (or are we just going with `.metal` for the config) but look for `.packet-cli` in its absence before prompting the `ConfigFileNotFoundError`. Relatively straightforward change as well, as we discussed.

go.mod
```diff
@@ -7,17 +7,15 @@ require (
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e // indirect
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
 	github.com/ghodss/yaml v1.0.0
+	github.com/golangci/golangci-lint v1.30.0 // indirect
```
displague: I think golangci-lint may have added several items to go.sum.

displague: I expect we can remove this line. I'm not sure how it was added. 
In some of the other projects we've found ways to get testing dependencies into a different go.mod file, but I don't think that's what is happening here.

displague:
> My plan currently is, also, to have the CLI prefer .metal-cli (or are we just going with .metal for the config

I proposed a different path in https://github.com/packethost/packet-cli/issues/54#issuecomment-848963718 (~/.config/equinix/metal.yaml). This would be an XDG compatible path and would fit into my desire to carve out `~/.equinix` as a location for numerous tools to share common configuration and extend configuration as needed (https://github.com/packethost/packngo/issues/189#issuecomment-802234842).

I haven't received much feedback on this plan. Despite the lack of feedback on those long-range goals, I think this is a reasonable fit for our immediate needs.

We appear to support a `--config` format for any users that would like to use an alternate config path. We could introduce an environment variable to override the path too(as a separate issue).

displague:
The README.md looks like it will need a once-over after applying the search/replace changes. 

For example, this will no longer be needed.

> Packet is now Equinix Metal!, keep an eye on this project for future project and command line name changes.


Documentation references to `PACKET_TOKEN` should be updated to `METAL_AUTH_TOKEN` (introduced in #119 ).

For the sake of 1.0, with the full CLI name change, we can remove support for the `PACKET_TOKEN` environment variable.  I feel the same way about the `.packet-cli.*` files, we have no reason to continue to support those names when the CLI tool has been renamed.

---

A way to hint to users that they do not have a valid configuration would be to emit a warning when no config file has been found.
`Warning: No configuration file was found. Use "metal init", "metal --config", or "METAL_AUTH_TOKEN=... metal"  to suppress this warning.`. We would then need a `metal init` tool to prompt for a token.  This could later be used to set the default organization and project (for #54).  `metal init` should be defined in a new issue. _Captured as https://github.com/packethost/packet-cli/issues/125_

jmarhee:
> The README.md looks like it will need a once-over after applying the search/replace changes.
> 
> For example, this will no longer be needed.
> 
> > Packet is now Equinix Metal!, keep an eye on this project for future project and command line name changes.
> 
> Documentation references to `PACKET_TOKEN` should be updated to `METAL_AUTH_TOKEN` (introduced in #119 ).
> 
> For the sake of 1.0, with the full CLI name change, we can remove support for the `PACKET_TOKEN` environment variable. I feel the same way about the `.packet-cli.*` files, we have no reason to continue to support those names when the CLI tool has been renamed.
> 
> A way to hint to users that they do not have a valid configuration would be to emit a warning when no config file has been found.
> `Warning: No configuration file was found. Use "metal init", "metal --config", or "METAL_AUTH_TOKEN=... metal" to suppress this warning.`. We would then need a `metal init` tool to prompt for a token. This could later be used to set the default organization and project (for #54). `metal init` should be defined in a new issue. _Captured as #125_

I'm going to look at these linked issues (previous comment as well), since all of this definitely seems worth doing here, especially if we're hoping to standardize; I didn't realize this had been proposed for other projects, and I agree that it's probably the way to go (rather than the patch I was going to propose in this PR for this)

jmarhee:
There's a couple more items to update to make the package rename complete, but I think it's most of the way there. The contribution guide, goreleaser.yml and release guide have Packet references left to do.

README.md
```diff
@@ -173,34 +171,34 @@ The Equinix Metal authentication token can be stored in the `$PACKET_TOKEN` envi
 Environment variable:
 
 ```bash
-export PACKET_TOKEN=[PACKET_TOKEN]
+export METAL_AUTH_TOKEN=[METAL_AUTH_TOKEN]
 ```
 
-YAML configuration file - `$HOME/.packet-cli.yaml`:
+YAML configuration file - `$HOME/.config/equinix/.metal.yaml`:
```
displague: Within XDG, this doesn't have to be _hidden_, this should be `metal.yaml` (no dot). 
```suggestion
YAML configuration file - `$HOME/.config/equinix/metal.yaml`:
```

cmd/root.go
```diff
@@ -114,18 +114,19 @@ func initConfig() {
 		// Use config file from the flag.
 		viper.SetConfigFile(cfgFile)
 	} else {
-		viper.SetConfigName(".packet-cli")
-		viper.AddConfigPath(userHomeDir())
+		configDir := userHomeDir() + "/.config/equinix"
+		viper.SetConfigName(".metal")
```
displague:
```suggestion
		viper.SetConfigName("metal")
```

jmarhee:
Noticed that, with the TOKEN environment variable unset, if the path to the config is changed (as we've done here), the CLI seems no longer able to read in the new config after I build-all; looking into this. I suspect it's a local issue, but will try to reproduce.

README.md
```diff
@@ -174,15 +174,15 @@ Environment variable:
...
+YAML configuration file - `$HOME/.config/equinix/metal.yaml`:
 
 ```yaml
 ---
 token: METAL_AUTH_TOKEN
 
 ```
 
-JSON configuration file - `$HOME/.config/equinix/metal.json`:
+JSON configuration file - `$HOME/.packet-cli.json`:
```
displague:
```suggestion
JSON configuration file - `$HOME/.config/equinix/metal.json`:
```

displague:
> Noticed that, with the TOKEN environment variable unset, if the path to the config is changed (as we've done here), the CLI seems no longer able to read in the new config after I build-all; looking into this. I suspect it's a local issue, but will try to reproduce.

I noticed this with the current `packet` binaries. The environment variable should override the config file settings, but currently, the config file overrides the environment variable. We can address this in a separate issue https://github.com/packethost/packet-cli/issues/126.

go.mod
```diff
@@ -1,23 +1,15 @@
...
-	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
 	github.com/ghodss/yaml v1.0.0
-	github.com/juju/ansiterm v0.0.0-20180109212912-720a0952cc2a // indirect
-	github.com/lunixbochs/vtclean v0.0.0-20170504063817-d14193dfc626 // indirect
 	github.com/manifoldco/promptui v0.3.0
-	github.com/mattn/go-colorable v0.0.9 // indirect
-	github.com/mattn/go-isatty v0.0.3 // indirect
-	github.com/mattn/go-runewidth v0.0.2 // indirect
 	github.com/olekukonko/tablewriter v0.0.0-20180506121414-d4647c9c7a84
+	github.com/packethost/packet-cli v0.0.0-20210527020038-319668ff3c78
```
displague: This is a sign that `packethost/packet-cli` (rather than `equinix/metal-cli`) is still be used in one (some?) of the imports.

cmd/root.go
```diff
@@ -114,18 +114,19 @@ func initConfig() {
 		// Use config file from the flag.
 		viper.SetConfigFile(cfgFile)
 	} else {
-		viper.SetConfigName(".packet-cli")
-		viper.AddConfigPath(userHomeDir())
+		configDir := userHomeDir() + "/.config/equinix"
```
displague: perhaps we should use `path.Join` here.

Makefile
```diff
@@ -54,7 +54,7 @@ clean-docs:
 
 install:
 	go install ${LDFLAGS}
-	mv ${GOPATH}/bin/packet-cli ${GOPATH}/bin/packet
+	mv ${GOPATH}/bin/${BINARY} ${GOPATH}/bin/${BINARY}
```
displague:
```
$ make install
go install -ldflags "-X github.com/equinix/metal-cli/cmd.Version=f698724 -X github.com/equinix/metal-cli/cmd.Build=`git rev-parse --short HEAD`"
mv /Users/marques/.local/share/go/bin/metal /Users/marques/.local/share/go/bin/metal
mv: rename /Users/marques/.local/share/go/bin/metal to /Users/marques/.local/share/go/bin/metal: No such file or directory
make: *** [install] Error 1
```


```suggestion
	mv ${GOPATH}/bin/metal-cli ${GOPATH}/bin/${BINARY}
```

Perhaps we can create a new variable (`PROJECT_NAME`?) that is used by `PACKAGE_NAME` and this `mv` command.

displague (changes requested):
This looks close, @jmarhee!

I left a comment at https://github.com/packethost/packet-cli/pull/124#discussion_r643502300 about `make install` not working.

I also see some unexpected results in:
* `git grep PACKET`
  * We should open a new issue to remove `PACKET_TOKEN`, and use `METAL_AUTH_TOKEN` in the github workflow
* `git grep packet`
  * Each cmd has a help description line which still contains the word `packet`. These should be updated to `metal`.  We can use a variable here if we change all of those blocks of text to a function that has the binary name filled in through go templating, but I think we're safe to simply replace `packet` with `metal` in these `cmd/` files for now.  We'll need to `make generate-docs` again.
  * There are some references to the `packet` binary name in the CI that all appear safe to rename to `metal`.
  * There are a few more references to `packethost/packet-cli`, including a reference in goreleaser. These should all be `equinix/metal-cli`.
* `git grep Packet`
  * > `[Packet is now Equinix Metal!](https://blog.equinix.com/blog/2020/10/06/equinix-metal-metal-and-more/), keep an eye on this project for future project and command line name changes.`
  
  We can remove the ",keep an eye on ..." because this is the rename PR!

displague (approved):

PR #130, opened 2021-06-22
Change Config vs Env precedence

### LINKED ISSUE #126: METAL_AUTH_TOKEN should override the config file settings
### What happened?

With `METAL_AUTH_TOKEN=foo packet ...` I would expect the environment variable to override the token defined in the .packet.yaml config file.

This is not the case today.

### How can we reproduce it?

* define a valid token in the config file
* define an invalid token in the environment variable
* the command should fail but it succeeds

> CLI version (`packet --version`):
v0.5.0

displague:
Fixes #126 

Base branch is `metal`, changes will be part of the _new_ metal-cli.

This PR gives `METAL_AUTH_TOKEN` precedence over `token` set in `~/.config/equinix/metal.yaml`. Previously, the environment variable was only used if no config value was set.

`PACKET_TOKEN` support is dropped.

Viper is used to fetch environment variables and config file variables.  These changes come from #90 and will allow for a cleaner PR to solve #54.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,6 @@ jobs:
     - name: Detect formatting and build errors
       run: make
       env:
-        PACKET_TOKEN: bogus
+        METAL_AUTH_TOKEN: bogus
     - name: Detect Uncommitted Docs
       run: git diff --exit-code docs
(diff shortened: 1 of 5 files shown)


displague:
Verified expected behavior, per the PR description, with `go run ./ env` `go run ./ projects get` and `go run ./ docs`.

I'm merging this now and deferring deeper review to #128.

PR #133, opened 2021-06-24
Refactor scope of all commands to avoid globals and keep CLI API internal

### LINKED ISSUE #33: Variable scoping is a bit concerning

displague:
Fixes #33 

Closes #66


displague:
Deferring deeper review to #128

PR #135, opened 2021-07-07
select default project id from config

### LINKED ISSUE #54: Offer a way to set defaults using the config file

displague:
Replaces #90

> This experimental branch will introduce the ability to define a default organization id and project id from the Packet CLI config file.
>
>  Project IDs are not memorable. It would be convenient to be able to packet device create with the fewest arguments possible.  The default facility and plan could be stored in the Packet CLI config, for example.
>
> The interplay of default values required arguments, and impliable arguments has the potential to complicate and simplify the UX.
>
> This is similar behavior to what one can expect from the Packet Web portal.
> Each user account has a default project and organization defined, so the values in the config file could be derived from the Packet API, or could fall back to those values when not provided in other commands.

Fixes #54

### DIFF
--- a/cmd/cli.go
+++ b/cmd/cli.go
@@ -49,7 +49,6 @@ func NewCli() *Cli {
 	}
 
 	rootClient := root.NewClient(consumerToken, apiURL, Version)
-	rootClient.Init()
 	rootCmd := rootClient.NewCommand()
 	rootCmd.DisableSuggestions = false
 	cli.MainCmd = rootCmd
--- a/docs/metal.md
+++ b/docs/metal.md
@@ -15,6 +15,7 @@ Command line interface for Equinix Metal
       --include strings   Comma seperated Href references to expand in results, may be dotted three levels deep
   -j, --json              JSON output
       --search string     Search keyword for use in 'get' actions. Search is not supported by all resources.
+      --token string      Metal API Token (METAL_AUTH_TOKEN)
   -y, --yaml              YAML output
 ```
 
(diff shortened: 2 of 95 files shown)


displague:
project-id now defaults to the environment variable `METAL_PROJECT_ID` or the `project-id` config setting.

Commands that _should_ not default to the preferred project id (such as `project get`) use a different flag (`id`, for example).

PR #137, opened 2021-07-16
reserve short flags for common use

### LINKED ISSUE #111: Reconsider the short flag letter assignments
Remove or replace conflicting single letter options

Some of the short options were added to hastily. These options are conflicting with other meanings. Certain short options should have universal meaning, while some existing universal options were unfortunately chosen.

Opinionated examples:
| Letter | Reason
| --- | ---
|`-i` | should be reserved for id, everywhere
|`-p` |should be reserved for project, everywhere
|`-f` |should mean force, everywhere. It means `facility` in many places.
|`-y`| could imply ("yes") for a force prompt. It means `yaml` universally.
|`-j` |is the counterpart to `-y`, the `--json` to `--yaml`.  `-j` should follow any changes to `-y`.

## Questions
* Should we adopt `-o yaml`, `-o json`, following `kubectl`? This would allow for more formats in the future.
* Should `-f` continue to mean `--facility`, and `--force` must always use the long word?
* Do we every want to use `-y` for yes?
* What other flags are used inconsistently?

displague:
Fixes #111 

--output json|yaml (-o) replaces --json/-j and --yaml/-y

-O replaces --organization-id and --operating-system short flag

My thinking is that capitalized flags can have command-specific meanings and we can use the obvious letter choice without colliding with global flags.

-i is reserved for --id, requiring a change to -I for --ipxe-script-url and -s for --spot-instance (previously -I)

-p is reserved for --project, capacity check --plan is now -P


displague:
I was considering changing `-f` (facility) to `-F`, to make way for `--force` in the future.  
I avoided this because `--metro` should perhaps become `-M` and I'm not ready to consider that change.

We only need `-y` (`--yes`) at the moment, to approve deletions, and have no immediate need for `--force` short flags.  This can be added in the future.

Where delete commands take a `force: true` API argument we can accept `--force` with no short flag.

PR #140, opened 2021-07-21
fix typo in go get command

displague:
Wrong URL in the install instructions.


displague:
I'm going to use this PR to address #141.

Go 1.16 recommends users use `go install module@version`, so we should update the instructions to match. https://blog.golang.org/go116-module-changes#TOC_4.

One problem I noticed with this approach is that the tagged version is not reflected in `metal --version`.

We should reflect the build version, as was done here: https://github.com/packethost/packngo/blob/master/version.go

displague:
Deferring the above comments to #143

PR #143, opened 2021-07-23
use runtime detected version for go install binaries

displague:
When installed via `go install`, the reported version is `devel`. This should report the version specified in the go install arguments.

_Originally https://github.com/equinix/metal-cli/pull/140#issuecomment-884948534_


displague:
This didn't give the desired result.
```
$ go install github.com/equinix/metal-cli/cmd/metal@0.6.0-alpha2
$ metal --version
metal version v0.0.0-20210723042849-acf96ea52a20
```

Same result for `@latest`.

PR #150, opened 2021-07-24
go/ci/makefile: update to use go 1.16 and fixes on makefile and ci files

cpanato:
- upgrade project to go 1.16
- small fixes in the makefile
- update CI to use run on main (master branch does not exist anymore) and to use go 1.16
- update golangci


displague:
Perhaps we should use a util/go.mod (some sub-directory) for golangci-lint to avoid the dependency noise.

cpanato:
@displague did some changes and to hold the golangci-lint I did a similar approach we have for cluster-api projects and that does not impact in the go modules for the project and we also don't need to maintain another go module.

If you don't like this and prefer having another go module for tools let me know and I will revert and use the extra go module approach.

displague:
Thanks, @cpanato!

PR #155, opened 2021-10-06
add sort-by and sort-dir global get arguments

displague:
Signed-off-by: Marques Johansson <mjohansson@equinix.com>


displague:
Needs to include docs

PR #156, opened 2021-10-06
add "ssh-key get --project-id $ID"

displague:
This PR add support for listing ssh-keys in a given project.

```
metal ssh-keys get -P 
```

### DIFF
--- a/docs/metal_ssh-key_get.md
+++ b/docs/metal_ssh-key_get.md
@@ -7,7 +7,7 @@ Retrieves a list of SSH keys or a specified SSH key.
 Retrieves a list of SSH keys associated with the current user's account or the details of single SSH key.
 
 ```
-metal ssh-key get [-i <SSH-key_UUID>] [flags]
+metal ssh-key get [-i <SSH-key_UUID>] [-P] [-p <project_id>] [flags]
 ```
 
 ### Examples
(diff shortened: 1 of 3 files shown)


displague:
Previous description:

> This PR, as drafted, introduces a bug because the project-id from ~/.config/equinix/metal.yaml will always be defined. 
> **A user will therefor not be able to get _user_ ssh-keys, they can only get _project_ ssh keys.**
>
> Stated another way, project id is defined by viper from config args. This is preventing the `ssh-key get` from defining project id.
> 
> Possible solutions:
> * use a different variable name for project selection (`metal ssh-key get --project --project-id {id}`)
> * use a different action name for user or project ssh key listing (`metal project ssh-keys`, `metal ssh-key project-get`)
> * default to project listing and only attempt user ssh-key listing in the presence of a new argument (`metal ssh-key get --user`)
> * find a way to ignore the project-id defined by viper (`metal ssh-key get --project-id $id` (ignoring the viper defined project-id))

---

I went with the first option, but expanded `--project` to `--project-ssh-keys` (with short flag `-P`):
`metal ssh-key get --project-ssh-keys --project-id {id}`

Since `metal init` sets a default project, `metal ssh-keys get -P` will list project ssh keys for the configured project.

PR #164, opened 2022-01-12
Add homebrew release to goreleaser 

### LINKED ISSUE #38: Install via brew?
Can we make the packet cli available on Mac (and Linux) via [homebrew](https://brew.sh)? Also might want to think about Windows, where you can install via homebrew using WSL, or possibly [Chocolatey](https://chocolatey.org).

crayzeigh:
Resolves #38 by adding a `brews:` section to `goreleaser.yml` to generate the updated homebrew formula with each release run.

Tested this locally by publishing to [crayzeigh/homebrew-tap](https://github.com/crayzeigh/homebrew-tap) using my fork and a fake 0.7.1 release. After running `goreleaser release` I could tap and install via homebrew successfully shown here:

install via homebrew

As written, the homebrew formula will be committed back to equinix/homebrew-tap by me. We can adjust either or both the tap or the commit author (default is `goreleaserbot (goreleaser@carlosbecker.com)`). 

I think if this works as the main method of release, the next step would be to get the formula accepted into homebrew-core and figure out the best way to maintain that on each metal-cli release


jeremytanner:
Nice. Time to `brew`!

PR #167, opened 2022-01-14
Add parsing of metros, faclities, and plans

cprivitere:
Signed-off-by: Chris Privitere <cprivite@users.noreply.github.com>

### DIFF
--- a/internal/capacity/retrieve.go
+++ b/internal/capacity/retrieve.go
@@ -21,33 +21,52 @@
 package capacity
 
 import (
+	"github.com/packethost/packngo"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 func (c *Client) Retrieve() *cobra.Command {
 	var (
-		checkFacility bool
-		checkMetro    bool
+		metros, facilities, plans, locs []string
+		metro, facility, plan           string
 	)
 	// retrieveCapacitiesCmd represents the retrieveCapacity command
 	var retrieveCapacityCmd = &cobra.Command{
-		Use:     "get",
+		Use:     `get {-m [metros,...] | -f [facilities,...]} -P [plans,...]`,
 		Aliases: []string{"list"},
-		Short:   "Returns a list of facilities or metros and plans with their current capacity.",
-		Long: `Example:
-Retrieve capacities:
-metal capacity get { --metro | --facility }
-`,
-		RunE: func(cmd *cobra.Command, args []string) error {
-			cmd.SilenceUsage = true
-			var err error
-			lister := c.Service.List
-			fieldName := "Facility"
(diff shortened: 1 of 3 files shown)


internal/capacity/retrieve.go
```diff
@@ -21,55 +21,115 @@
...
-			requiredDataFormat := [][]string{}
+			header := []string{locationField, "Plan", "Level"}
+			data := [][]string{}
 
 			for locCode, capacity := range *capacities {
 				for plan, bm := range capacity {
-					loc := []string{}
-					loc = append(loc, locCode, plan, bm.Level)
-					requiredDataFormat = append(requiredDataFormat, loc)
+					if len(locs) > 0 {
```
displague: We can move the `loc` filtering to the previous loop, we can `continue` from the `locCode` loop if `locCode` is not in the `locs` list.

It may help to create a `contains` helper function rather than to create the comparison loop twice.

displague (changes requested):
While testing this I discovered that the json/yaml output is not affected by the client-side filtering.

The following addresses that and reduces the depth of the for/if blocks.


```patch
diff --git a/internal/capacity/retrieve.go b/internal/capacity/retrieve.go
index 83cedfb..cdb7892 100644
--- a/internal/capacity/retrieve.go
+++ b/internal/capacity/retrieve.go
@@ -75,21 +75,26 @@ func (c *Client) Retrieve() *cobra.Command {
                        header := []string{locationField, "Plan", "Level"}
                        data := [][]string{}

+                       filtered := map[string]map[string]map[string]string{}
+
                        for locCode, capacity := range *capacities {
-                               // If the list of locations isn't empty and contains this location code
-                               if !(len(locs) > 0 && !contains(locs, locCode)) {
-                                       for plan, bm := range capacity {
-                                               loc := []string{}
-                                               // If the list of plans isn't empty and contains this plan
-                                               if !(len(plans) > 0 && !contains(plans, plan)) {
-                                                       loc = append(loc, locCode, plan, bm.Level)
-                                                       data = append(data, loc)
-                                               }
+                               if len(locs) > 0 && !contains(locs, locCode) {
+                                       continue
+                               }
+                               for plan, bm := range capacity {
+                                       if len(plans) > 0 && !contains(plans, plan) {
+                                               continue
+                                       }
+                                       loc := []string{locCode, plan, bm.Level}
+                                       data = append(data, loc)
+                                       if len(filtered[locCode]) == 0 {
+                                               filtered[locCode] = map[string]map[string]string{}
                                        }
+                                       filtered[locCode][plan] = map[string]string{"levels": bm.Level}
                                }
                        }

-                       return c.Out.Output(capacities, header, &data)
+                       return c.Out.Output(filtered, header, &data)
                },
        }
```

I'm sure we could further optimize this, we only need filtered in json/yaml mode, but I think this is well enough.

displague (approved):

PR #168, opened 2022-02-17
Updating the Docs - VLANs

DailyAlice:
- standardizes `Use:` field for each command
- adds content for both `Short:` and `Long:` fields in each command
- adds `Example:` commands where appropriate
- fixes typo in the global flags
- updates some of the flag descriptions to be a bit more explicit. Especially for --project-id.

### DIFF
--- a/internal/vlan/retrieve.go
+++ b/internal/vlan/retrieve.go
@@ -35,7 +35,7 @@ func (c *Client) Retrieve() *cobra.Command {
 		Use: `get -p <project_UUID>`,
 		Aliases: []string{"list"},
 		Short: "Lists virtual networks.",
-		Long: "Retrieves a list of virtual networks for the specified project.",
+		Long: "Retrieves a list of all VLANs for the specified project.",
 		Example:  `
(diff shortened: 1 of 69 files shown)


internal/vlan/retrieve.go
```diff
@@ -32,14 +32,14 @@ func (c *Client) Retrieve() *cobra.Command {
...
-		Long: `Example:
-
-metal virtual-network get -p [project_UUID]
-
-	`,
+		Short: "Lists virtual networks.",
+		Long: "Retrieves a list of all VLANs for the specified project.",
+		Example:  `
+  # Lists virtual networks for project 3b0795ba-ec9a-4a9e-83a7-043e7e11407c:
+  virtual-network get -p 3b0795ba-ec9a-4a9e-83a7-043e7e11407c`,
```
cprivitere: Should this say "metal virtual-network get -p 3b0795ba-ec9a-4a9e-83a7-043e7e11407c"?

DailyAlice: yes, good catch.

displague:
Thanks for these improvements! Thanks for seperating your commits!

PR #170, opened 2022-02-22
Updates docs for users, two-factor, and ssh commands

DailyAlice:
- Updates the man text with use, short, long and examples.
- Ran the docs build command to update the markdown files.


cprivitere (approved):
LGTM

PR #171, opened 2022-02-23
Docs updates to the project and plan commands

DailyAlice:
- adds/updates Use, Short, Long, and Example fields for the project and plan commands
- builds and updates the markdown docs for both


cprivitere (approved):
LGTM

PR #175, opened 2022-03-10
Updating docs for the env and docs commands

DailyAlice:
- Updates use, short, long, and example fields for both commands 
- Updates the generated markdown docs

Note: might want to have some eyes on the `env` updates specifically, I'm not super confident I captured the right things.


cprivitere (commented):
So this mostly looks good but it's making me wonder if the eval function even needs the -p option anymore.

PR #179, opened 2022-03-18
Add an emdocs command to generate a single page doc

displague:
adds an emdocs command to generate a single page doc.

This is particular interest to @DailyAlice


DailyAlice:
So I am happy with the format and generation of the emdocs page. Not too sure what the merge plan is for this PR, and I'm also not too sure how to make it pass the tests with the additional markdown file.

Thoughts @displague and @cprivite?

docs/emdocs.md
```diff
@@ -0,0 +1,2379 @@
+### metal 2fa
```
displague: I think we don't need to include emdocs.md here. I would imagine any external processes that require this file will run the `metal emdocs` command in the environment where emdocs.md is needed.

PR #183, opened 2022-03-23
`packet operating-systems get` does not return a list in a sorted order

### LINKED ISSUE #78: `packet operating-systems get` does not return a list in a sorted order
### What happened?

I'm trying to figure out exactly how to provision a particular operating
system, and to determine what my choices are.

`packet-cli operating-systems get` looks like it's the right command,
but the list is very long, and it's not sorted, and I know that only 
some combination of operating systems and versions are available for
some machines.

At minimum it seems like the list should be returned in a stable, alphabetically
sorted order.

### How can we reproduce it?

`packet-cli operating-systems get`

CLI version (`packet --version`): packet version 0.0.8

cprivitere:
Fixes #78

Signed-off-by: Chris Privitere <cprivite@users.noreply.github.com>

### DIFF
--- a/internal/os/retrieve.go
+++ b/internal/os/retrieve.go
@@ -21,16 +21,18 @@
 package os
 
 import (
+	"sort"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 func (c *Client) Retrieve() *cobra.Command {
 	return &cobra.Command{
-		Use: `get`,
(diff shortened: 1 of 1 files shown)


displague:
This LGTM as a fix for #78.

Long term it would help to find a pattern that doesn't require one-off changes to each command or arbitrary (and hardcoded) field names.

Issue #185, opened 2022-03-24
Add binary release for darwin/arm64


vielmetti:
### What problem are you facing?

No binary image is made available for darwin/arm64, which means that installation of the `metal-cli` on an Apple M1 Macintosh is harder than necessary.

### How could the Equinix Metal CLI help solve this problem?

Update the "goreleaser" configuration to cause it to generate darwin/arm64 binaries.

vielmetti:
cc @npmccallum

vielmetti:
Fixed in https://github.com/equinix/metal-cli/pull/186

Issue #184, opened 2022-03-24
Install via Homebrew fails for M1 Apple Macintosh


vielmetti:
### What happened?

`metal-cli` fails to install on M1 Apple Macintosh using the homebrew tap.

### How can we reproduce it?

brew tap equinix/homebrew-tap

---

Report relayed from customer:

I am on an M1 mac. I tried installing metal-cli, but failed.

1. There is no pre-build binary for this platform.
2. Attempting to install via brew produces this error:
brew tap equinix/homebrew-tap

==> Tapping equinix/tap

Cloning into '/opt/homebrew/Library/Taps/equinix/homebrew-tap'...

remote: Enumerating objects: 26, done.

remote: Counting objects: 100% (18/18), done.

remote: Compressing objects: 100% (15/15), done.

remote: Total 26 (delta 8), reused 6 (delta 2), pack-reused 8

Receiving objects: 100% (26/26), 6.87 KiB | 6.87 MiB/s, done.

Resolving deltas: 100% (8/8), done.

Error: Invalid formula: /opt/homebrew/Library/Taps/equinix/homebrew-tap/metal-cli.rb

formulae require at least a URL

Error: Cannot tap equinix/tap: invalid syntax in tap!

vielmetti:
It's likely that #185 will need to be addressed as part of this as well.

vielmetti:
I was able to confirm this issue on my own machine.

Note that installation using "go install" does work just fine, 

`go install github.com/equinix/metal-cli/cmd/metal@latest`

per current directions.

cc @npmccallum

vielmetti:
Fixed in https://github.com/equinix/metal-cli/pull/186

PR #186, opened 2022-03-24
removes build ignore for windows/darwin arm

crayzeigh:
Removes the ignore lines from goreleaser so builds will continue for windows and darwin arm instances and publish artifacts.


crayzeigh:
Fixes #185 and #184

PR #218, opened 2022-09-20
ensure IP lists can be filtered (types=public_ipv6)

displague:
Ensure that the IP lists can be filtered using the available query params.
```
metal ips list --filter types=public_ipv4
metal ips list --filter types=public_ipv6
```
... and others


ocobles (approved):
useful for --filter and also --include --exclude options

PR #220, opened 2022-09-27
add "metal user add" command

### LINKED ISSUE #219: metal add user
### What problem are you facing?
I want to add users as admins or collaborators to one or more projects with metal-cli.

### How could the Equinix Metal CLI help solve this problem?
`metal user add foo@example.com` with optional `-p proj_id1 -p proj_id2 -p proj_id3`

Because `metal init` ensures the default project is used, I'd like to get away with using the bare statement above for the majority of use cases.

I’ll need a `--roles` option too (or multiple --role arguments).

displague:
Fixes #219

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.16
 require (
 	github.com/manifoldco/promptui v0.9.0
 	github.com/olekukonko/tablewriter v0.0.5
-	github.com/packethost/packngo v0.26.0
+	github.com/packethost/packngo v0.27.0
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.5.0
 	github.com/spf13/pflag v1.0.5
--- a/internal/users/add.go
+++ b/internal/users/add.go
@@ -1,0 +1,77 @@
+// Copyright © 2022 Equinix Metal Developers <support@equinixmetal.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -5,7 +5,8 @@ go 1.16
 require (
 	github.com/manifoldco/promptui v0.9.0
 	github.com/olekukonko/tablewriter v0.0.5
-	github.com/packethost/packngo v0.26.0
+	// TODO: do not merge this. remove this once https://github.com/packethost/packngo/pull/346 is merged
+	github.com/packethost/packngo v0.26.1-0.20220926212055-cc3c8410dbe8 // indirect
```
displague: Do not merge without addressing this first. Depends on packethost/packngo#346

internal/users/add.go
```diff
@@ -0,0 +1,77 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
+			getOpts := c.Servicer.ListOptions([]string{"invited_by", "invitable", "projects"}, nil)
+			createRequest := &packngo.InvitationCreateRequest{
+				Invitee:        email,
+				Roles:          roles,
+				ProjectIDs:     projectIDs,
+				OrganizationID: organizationID,
+			}
+			invitation, _, err := c.InvitationService.Create(createRequest, getOpts)
```
displague: I'm having trouble identifying why this API call (with a program id, role, organization id, and invitee) is resulting in:
```
{
 "errors": [
  "Roles contains an invalid role limited_collaborator"
 ]
}
```

Making the same request through the web portal (and investigating the network panel) reveals that the requests are identical.  The differences that I feel need to be further scrutinized for this PR, are the HTTP headers.

displague: `project_ids` was supposed to be `projects_ids` 🤦

ocobles (approved):
lgtm

PR #233, opened 2022-10-25
Add support for device reinstall

mmlb:
Signed-off-by: Manuel Mendez <github@i.m.mmlb.dev>


internal/devices/reinstall.go
```diff
@@ -0,0 +1,65 @@
...
+	)
+
+	reinstallDeviceCmd := &cobra.Command{
+		Use:   `reinstall -d <device-id>`,
+		Short: "Reinstalls a device.",
+		Long:  "Reinstalls the provided device. The ID of the device to reinstall is required.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			request := packngo.DeviceReinstallFields{
+				OperatingSystem: "",
```
displague, line 45: It looks like we missed the `operatingSystem` here and reused `preserveData` for `DeprovisionFast` (🤔 maybe that combination makes sense?).  There are a few other flags we could have exposed.

mmlb: Should be fixed by https://github.com/equinix/metal-cli/pull/238. PreserveData and DeprovisionFast are tied to the same option because it doesn't make sense to have one without the other. osie will only skip wiping disks if both are set.

PR #238, opened 2022-11-10
reinstall: pass operatingSystem and split PreserveData and DeprovisionFast

mmlb:
I forgot to pass `operatingSystem` in to the packngo call.

I've changed the code so `PreseveData` and `FastDeprovision` aren't tied together. In the backend deprovision process `PreserveData` currently implies `DeprovisionFast`. `DeprovisionFast` is currently only useful to avoid the slow wipe of disks that will be used to reinstall the os into. In the future `DeprovisionFast` could also skip lengthy inventory collection steps that don't really need to happen for a reinstall.

### DIFF
--- a/internal/devices/reinstall.go
+++ b/internal/devices/reinstall.go
@@ -32,6 +32,7 @@ func (c *Client) Reinstall() *cobra.Command {
 		id string
 
 		operatingSystem string
+		deprovisionFast bool
 		preserveData    bool
 	)
 
@@ -44,7 +45,7 @@ func (c *Client) Reinstall() *cobra.Command {
 			request := packngo.DeviceReinstallFields{
 				OperatingSystem: operatingSystem,
 				PreserveData:    preserveData,
-				DeprovisionFast: preserveData,
+				DeprovisionFast: deprovisionFast,
 			}
 
 			_, err := c.Service.Reinstall(id, &request)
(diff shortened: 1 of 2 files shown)


internal/devices/reinstall.go
```diff
@@ -42,7 +42,7 @@ func (c *Client) Reinstall() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 
 			request := packngo.DeviceReinstallFields{
-				OperatingSystem: "",
+				OperatingSystem: operatingSystem,
 				PreserveData:    preserveData,
 				DeprovisionFast: preserveData,
```
displague: Does it make sense to pin `DeprovisionFast` to `preserveData`?
What are the benefits of preserving data without deprovisioning fast?

mmlb: I had this reasoning in the wrong place... in a comment on the previous pr https://github.com/equinix/metal-cli/pull/233/files#r1019605111. I've changed the code so they aren't tied together. Its actually the other way around. `PreserverData` currently implies `DeprovisionFast` in the deprovision process, `DeprovisionFast` currently is only useful to avoid the slow wipe of disks that need it. In the future `DeprovisionFast` could also skip lengthy inventory collection steps that don't really need to happen for a reinstall.

displague (approved):

PR #239, opened 2022-12-02
correct hardware-reservation get/move examples to use -

cturra:
This is a small update to correct a few `hardware-reservation` `get` and `move` examples to swap a `-` instead of `_`.

Thanks for this tool, it's been very useful 🙏🏼


DailyAlice (approved):
LGTM, thank you.

PR #262, opened 2023-04-27
fix: updated API token portal path

codinja1188:
we are updated the API token portal path.

### DIFF
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -76,8 +76,8 @@ func (c *Client) NewCommand() *cobra.Command {
 				config = c.Servicer.DefaultConfig(true)
 			}
 
-			fmt.Print("Equinix Metal API Tokens can be obtained through the portal at https://console.equinix.com/.\nSee https://metal.equinix.com/developers/docs/accounts/users/ for more details.\n\n")
-			fmt.Print("Token (hidden): ")
+			fmt.Print("Equinix Metal API Tokens can be obtained through the portal at https://console.equinix.com/profile/api-keys\nSee https://metal.equinix.com/developers/docs/accounts/users/ for more details.\n\n")
+			fmt.Print("User Token (hidden): ")
 			b, err := term.ReadPassword(int(syscall.Stdin))
 			if err != nil {
 				return err


codinja1188:
vasu@vasu-cloud:~/Equinix/metal-cli$ metal init
Equinix Metal API Tokens can be obtained through the portal at https://console.equinix.com/profile/api-keys
See https://metal.equinix.com/developers/docs/accounts/users/ for more details.

User Token (hidden): 
Organization ID []: 
Project ID []:

ctreatma:
I think `metal-cli` is expected to support user tokens and project tokens (someone else should correct me if I'm wrong).  Rather than updating the console URL in the help text to point at the user token page, we may want to update the docs URL to point at a document that explains the 2 token types.

codinja1188:
> I think `metal-cli` is expected to support user tokens and project tokens (someone else should correct me if I'm wrong). Rather than updating the console URL in the help text to point at the user token page, we may want to update the docs URL to point at a document that explains the 2 token types.

"metal init" always looking for user API keys only right?

displague:
> "metal init" always looking for user API keys only right?

`metal init` should work with either a Project or a User API token. If you use a project API token you won't be able to perform certain actions, but that may be desirable by the user.

One of the actions `metal init` performs is a User Preferences check to determine the default organization and project. That won't work with a project API token.

I'd propose an alternative lookup process during `metal init`:

- GET /projects with the token provided. limit the results to 2 (we only care if the count is == 1)
  - if there is an error, the token is invalid (ask again)
  - if there are 0, warn and skip project detection. (I don't believe metal-cli should offer to create an initial project for the user)
  - If there is 1 project:
    - just use that project as the default. we will also know the default organization by taking it from that project. (this is new code, iirc)
  - otherwise, we know this is a User API Key and we'll need to know which project should be default.

- fetch the User Preferences to discover the default project and org (this is existing code)
  - prompt the user with the project defaulted (this is existing code)

displague:
Broke that comment out into https://github.com/equinix/metal-cli/issues/265. This is good as-is and should prevent users from providing project API keys now, which would result in an error:
https://github.com/equinix/metal-cli/blob/main/internal/init/init.go#L92
```
Error: Could not get current User: GET https://api.equinix.com/metal/v1/user: 403 Access denied for the current authentication token
```

PR #264, opened 2023-05-12
feat: add --http-header flag for adding additional headers

0xch4z:
Adds the ability to add arbitrary http headers to requests

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -23,6 +23,7 @@ package cli
 import (
 	"fmt"
 	"log"
+	"net/http"
 	"os"
 	"path"
 	"runtime"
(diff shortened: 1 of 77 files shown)


displague:
Each action tends to call a single API endpoint, and perhaps this is always the case.
Because of this, I'm not worried about the user supplying a header intended for one API call and not other API calls within the context of the complete CLI transaction.

Update:
- `metal init` is an exception to this. Again, I think we'll be ok there for the known potential use-cases of `--http-header`.
- If we implement `--wait` to poll on resource creation until ready, we'll have actions that call multiple endpoints, repetitively

internal/cli/root.go
```diff
@@ -70,7 +85,7 @@ func NewClient(consumerToken, apiURL, Version string) *Client {
 }
 
 func (c *Client) apiConnect() error {
-	client, err := packngo.NewClientWithBaseURL(c.consumerToken, c.metalToken, nil, c.apiURL)
+	client, err := packngo.NewClientWithBaseURL(c.consumerToken, c.metalToken, c.httpClient, c.apiURL)
```
displague: I think we could reduce the plumbing a little if the headers were stored in the `Client` during `Config` rather than an `httpClient`. Or (better yet?)...

Storing the client in the config bothers at the surface level even though we are already doing it with the packngo.Client, but especially because we would be storing the httpClient in the config twice (within the packngo.Client and separately).


`Client.API` which is passed `cmd`, could
- if c.apiClient == nil, call `getAdditionalHeaders`
- create the httpClient
- pass it as an argument to apiConnect

PR #274, opened 2023-06-05
Added client metal-go

codinja1188:

This PR replaces packngo with metal-go for all interactions with the Equinix Metal API. The metal-go SDK is generated from the Equinix Metal API spec, and we will gradually replace all usage of packngo with metal-go to streamline support. we updated init with Metal-go client

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -29,6 +29,7 @@ import (
 	"runtime"
 	"strings"
 
+	metal "github.com/equinix-labs/metal-go/metal/v1"
 	"github.com/packethost/packngo"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -40,11 +41,13 @@ import (
 const (
 	envPrefix                  = "METAL"
 	configFileWithoutExtension = "metal"
+	debugVar                   = "PACKNGO_DEBUG"
 )
 
 type Client struct {
 	// apiClient client
-	apiClient *packngo.Client
+	apiClient      *packngo.Client
+	metalApiClient *metal.APIClient
 
 	includes      *[]string // nolint:unused
 	excludes      *[]string // nolint:unused
@@ -83,13 +86,30 @@ func NewClient(consumerToken, apiURL, Version string) *Client {
 	}
 }
 
+// This function provides backwards compatibility for the packngo
+// debug environment variable while allowing us to introduce a new
+// debug variable in the future that is not tied to packngo
+func checkEnvForDebug() bool {
+	return os.Getenv(debugVar) != ""
+}
+
 func (c *Client) apiConnect(httpClient *http.Client) error {
 	client, err := packngo.NewClientWithBaseURL(c.consumerToken, c.metalToken, httpClient, c.apiURL)
 	if err != nil {
-		return fmt.Errorf("Could not create Client: %w", err)
+		return fmt.Errorf("could not create client: %w", err)
 	}
 	client.UserAgent = fmt.Sprintf("metal-cli/%s %s", c.Version, client.UserAgent)
 	c.apiClient = client
+	return nil
+}
+
+func (c *Client) metalApiConnect(httpClient *http.Client) error {
+	configuration := metal.NewConfiguration()
+	configuration.Debug = checkEnvForDebug()
+	configuration.AddDefaultHeader("X-Auth-Token", c.Token())
+	configuration.UserAgent = fmt.Sprintf("metal-cli/%s %s", c.Version, configuration.UserAgent)
+	metalgoClient := metal.NewAPIClient(configuration)
+	c.metalApiClient = metalgoClient
 	return nil
 }
 
@@ -167,6 +187,26 @@ func (c *Client) API(cmd *cobra.Command) *packngo.Client {
 		}
 	}
 	return c.apiClient
+}
+
+func (c *Client) MetalAPI(cmd *cobra.Command) *metal.APIClient {
+	if c.metalToken == "" {
+		log.Fatal("Equinix Metal authentication token not provided. Please set the 'METAL_AUTH_TOKEN' environment variable or create a configuration file using 'metal init'.")
+	}
+
+	if c.metalApiClient == nil {
+		httpClient := &http.Client{
+			Transport: &headerTransport{
+				header: getAdditionalHeaders(cmd),
+			},
+		}
+
+		err := c.metalApiConnect(httpClient)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+	return c.metalApiClient
 }
 
 func (c *Client) Token() string {
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -22,12 +22,13 @@ THE SOFTWARE.
 package init
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 	"syscall"
 
-	"github.com/packethost/packngo"
+	metal "github.com/equinix-labs/metal-go/metal/v1"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
 	"sigs.k8s.io/yaml"
@@ -35,8 +36,8 @@ import (
 
 type Client struct {
 	Servicer       Servicer
-	UserService    packngo.UserService
-	ProjectService packngo.ProjectService
+	UserService    metal.UsersApiService
+	ProjectService metal.ProjectsApiService
 }
 
 func NewClient(s Servicer) *Client {
@@ -85,25 +86,26 @@ func (c *Client) NewCommand() *cobra.Command {
 			fmt.Println()
 			token := string(b)
 			c.Servicer.SetToken(token)
-			metalClient := c.Servicer.API(cmd)
-			c.UserService = metalClient.Users
-			c.ProjectService = metalClient.Projects
+			metalGoClient := c.Servicer.MetalAPI(cmd)
+			c.UserService = *metalGoClient.UsersApi
+			c.ProjectService = *metalGoClient.ProjectsApi
 
-			user, _, err := c.UserService.Current()
+			user, _, err := c.UserService.FindCurrentUser(context.Background()).Execute()
 			if err != nil {
 				return err
 			}
-			organization := user.DefaultOrganizationID
-			project := ""
-			if user.DefaultProjectID != nil {
-				project = *user.DefaultProjectID
-			}
+			organization := user.AdditionalProperties["default_organization_id"]
+			project := fmt.Sprintf("%v", user.AdditionalProperties["default_project_id"])
 			fmt.Printf("Organization ID [%s]: ", organization)
+
+			defaultOrganizationId := fmt.Sprintf("%v", organization)
 
 			userOrg := ""
 			fmt.Scanln(&userOrg)
 			if userOrg == "" {
-				userOrg = organization
+				if defaultOrganizationId != "" {
+					userOrg = defaultOrganizationId
+				}
 			}
 
 			// Choose the first project in the preferred org
@@ -113,6 +115,7 @@ func (c *Client) NewCommand() *cobra.Command {
 					return err
 				}
 			}
+
 			fmt.Printf("Project ID [%s]: ", project)
 
 			userProj := ""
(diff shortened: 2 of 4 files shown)


internal/cli/root.go
```diff
@@ -93,6 +95,15 @@ func (c *Client) apiConnect(httpClient *http.Client) error {
 	return nil
 }
 
+func (c *Client) metalApiConnect(httpClient *http.Client) error {
+	configuration := metal.NewConfiguration()
+	configuration.Debug = false
```
ctreatma: This should be set to true if the `PACKNGO_DEBUG` env var is provided.  Here's an example of how we could do that from the clusterapi provider: https://github.com/kubernetes-sigs/cluster-api-provider-packet/pull/527/files#diff-febb66857beae6d70b0aad1c73c92369afe1fb7d309f699f6047af7912d99e05R369-R374

internal/cli/root.go
```diff
@@ -93,6 +95,15 @@ func (c *Client) apiConnect(httpClient *http.Client) error {
...
+	configuration.AddDefaultHeader("X-Auth-Token", c.Token())
```
ctreatma, line 109: We should update the `configuration` to include `metal-cli` in the user-agent so that we maintain visibility into `metal-cli` usage:
```suggestion
	configuration.AddDefaultHeader("X-Auth-Token", c.Token())
	configuration.UserAgent = fmt.Sprintf(fmt.Sprintf("metal-cli/%s %s", c.Version, configuration.UserAgent)
```

internal/cli/root.go
```diff
@@ -169,6 +180,32 @@ func (c *Client) API(cmd *cobra.Command) *packngo.Client {
...
+
+		err := c.metalApiConnect(httpClient)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+	return c.metalApiClient
+}
+
+// func getMetalVersion() string {
```
ctreatma: This can be removed rather than commenting out

internal/cli/root.go
```diff
@@ -220,6 +257,8 @@ func (c *Client) NewCommand() *cobra.Command {
 	rootCmd.PersistentFlags().StringVar(&c.sortDir, "sort-dir", "", "Sort field direction for use in 'get' actions. Sort is not supported by all resources.")
 
 	rootCmd.Version = c.Version
+
+	//rootCmd.Version = getMetalVersion()
```
ctreatma: This can also be removed rather than commented out

internal/init/init.go
```diff
@@ -85,40 +86,31 @@ func (c *Client) NewCommand() *cobra.Command {
...
-			}
+			organization := user.AdditionalProperties["default_organization_id"]
+
 			fmt.Printf("Organization ID [%s]: ", organization)
 
 			userOrg := ""
 			fmt.Scanln(&userOrg)
 			if userOrg == "" {
-				userOrg = organization
+				userOrg = fmt.Sprintf("value: %v", organization)
```
ctreatma: Is the `value: ` prefix needed?  I would expect this to break anything that uses the config file.

internal/init/init.go
```diff
@@ -85,40 +86,31 @@ func (c *Client) NewCommand() *cobra.Command {
...
-			}
-			fmt.Printf("Project ID [%s]: ", project)
+			project := user.AdditionalProperties["default_project_id"]
+			fmt.Printf("Default Project ID [%s]: ", project)
 
 			userProj := ""
 			fmt.Scanln(&userProj)
 			if userProj == "" {
-				userProj = project
+				userProj = fmt.Sprintf("value: %v", organization)
```
ctreatma: This should be `project`, not `organization`; as with the previous comment I'm not clear what the `value: ` prefix is for.

internal/cli/root.go
codinja1188: updated

internal/init/init.go
codinja1188: updated

internal/init/init.go
codinja1188: yes, ```value:``` is not needed, but ```metal init``` is going generate/write config file

internal/cli/root.go
```diff
@@ -219,7 +265,7 @@ func (c *Client) NewCommand() *cobra.Command {
 	rootCmd.PersistentFlags().StringVar(&c.sortBy, "sort-by", "", "Sort fields for use in 'get' actions. Sort is not supported by all resources.")
 	rootCmd.PersistentFlags().StringVar(&c.sortDir, "sort-dir", "", "Sort field direction for use in 'get' actions. Sort is not supported by all resources.")
 
-	rootCmd.Version = c.Version
+	rootCmd.Version = getMetalVersion()
```
ctreatma: This should still be `c.Version`; the `rootCmd.Version` is the version number for metal-cli, not for the library used by metal-cli.

internal/cli/root.go
```diff
@@ -169,6 +189,32 @@ func (c *Client) API(cmd *cobra.Command) *packngo.Client {
...
+
+		err := c.metalApiConnect(httpClient)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+	return c.metalApiClient
+}
+
+func getMetalVersion() string {
```
ctreatma: This can be removed; metal-cli doesn't need to know the version of the underlying SDK at runtime.

internal/cli/root.go
```diff
@@ -40,11 +41,13 @@ import (
 const (
 	envPrefix                  = "METAL"
 	configFileWithoutExtension = "metal"
+	DebugVar                   = "PACKNGO_DEBUG"
```
ctreatma: This should start with a lower-case letter (`debugVar`)

internal/init/init.go
ctreatma: The `value: ` prefix should be removed so that the config file is valid
```suggestion
				userOrg = organization
```

internal/init/init.go
ctreatma: The `value: ` prefix should be deleted and this should be set to a project ID so that the config file is valid.
```suggestion
				userProj = project
```

internal/init/init.go
```diff
@@ -170,8 +144,8 @@ func writeConfig(config string, b []byte) error {
 }
 
 type Servicer interface {
-	API(*cobra.Command) *packngo.Client
-	ListOptions(defaultIncludes, defaultExcludes []string) *packngo.ListOptions
+	MetalAPI(*cobra.Command) *metal.APIClient
```
ctreatma, line 196: Unless we _have_ to change this to `MetalAPI` to avoid errors, I'd say keep it as `API`
```suggestion
	API(*cobra.Command) *metal.APIClient
```

internal/init/init.go
```diff
@@ -170,8 +144,8 @@ func writeConfig(config string, b []byte) error {
...
+	//ListOptions(defaultIncludes, defaultExcludes []string) *packngo.ListOptions
```
ctreatma:
```suggestion
```

internal/init/init.go
codinja1188: It is shows following error "cannot use organization (variable of type interface{}) as string value in assignment: need type assertion"

internal/cli/root.go
codinja1188: updated

internal/init/init.go
codinja1188: As you know we are keeping packngo and metal-go in. we plan in seperate PR to rename MetalAPI -> API at the time we deprecated the packngo completely

internal/init/init.go
codinja1188: deleted

internal/cli/root.go
codinja1188: updated

internal/cli/root.go
codinja1188: deleted

internal/init/init.go
codinja1188: It is shows following error "cannot use organization (variable of type interface{}) as string value in assignment: need type assertion"

codinja1188: Here is the output
```
vasu@vasu-cloud:~/Equinix/metal-cli/bin$ cat /home/vasu/.config/equinix/metal.yaml
---
organization-id: *****************
project-id: ****************
token: *************
```

ctreatma: @vasubabu you will need to delete that API token in the Metal Console UI ASAP so that no one can use it and then come back and remove it from your comment as well.

codinja1188: it's not a original token. I will remove it

internal/init/init.go
ctreatma: Ah, right, this needs to match the function name in `root.go`

internal/init/init.go
```diff
@@ -85,40 +86,31 @@ func (c *Client) NewCommand() *cobra.Command {
...
-			// Choose the first project in the preferred org
-			if project == "" {
-				project, err = getFirstProjectID(c.ProjectService, userOrg)
-				if err != nil {
-					return err
-				}
-			}
-			fmt.Printf("Project ID [%s]: ", project)
+			project := user.AdditionalProperties["default_project_id"]
+			fmt.Printf("Default Project ID [%s]: ", project)
```
ctreatma: This prompt should not change; it's allowing the user to specify a project ID while also showing them what value will be used if they don't override it.
```suggestion
			fmt.Printf("Project ID [%s]: ", project)
```

internal/init/init.go
```diff
@@ -85,40 +86,31 @@ func (c *Client) NewCommand() *cobra.Command {
...
 			userOrg := ""
 			fmt.Scanln(&userOrg)
 			if userOrg == "" {
-				userOrg = organization
+				userOrg = fmt.Sprintf("%v", organization)
 			}
 
-			// Choose the first project in the preferred org
-			if project == "" {
-				project, err = getFirstProjectID(c.ProjectService, userOrg)
```
ctreatma: Are you sure we can remove this?  This code would look up the first project ID in the user's default organization if they have a non-nil default organization ID and a nil default project ID.  That situation should be possible regardless of whether we use packngo or metal-go

internal/init/init.go
```diff
@@ -85,40 +86,31 @@ func (c *Client) NewCommand() *cobra.Command {
...
+			c.ProjectService = *metalGoClient.ProjectsApi
 
-			user, _, err := c.UserService.Current()
+			user, _, err := c.UserService.FindCurrentUser(context.Background()).Execute()
 			if err != nil {
 				return err
 			}
-			organization := user.DefaultOrganizationID
-			project := ""
-			if user.DefaultProjectID != nil {
```
ctreatma, line 98: It doesn't look like we have a corresponding nil check in the new code; this is checking that the user's default project ID is not nil.

internal/init/init.go
codinja1188: updated

internal/init/init.go
codinja1188: I don't think, I feel at least one project should be default at any case, anyway we can get default Org ID using Metal-go

ctreatma: The user will have a default org, but the code that is being removed here looks for an alternative project ID if the user doesn't have a default project ID, and I think we need to keep that so that the init command behavior stays the same

codinja1188: yes, agree. we don't have relation ship  API to get Org ID from projects, We have API to get projects form Org ID

internal/init/init.go
codinja1188: updated

internal/init/init.go
```diff
@@ -132,18 +135,19 @@ func (c *Client) NewCommand() *cobra.Command {
...
-	projects, _, err := s.List(listOpts)
+	projects, _, err := s.FindProjects(context.Background()).Include(include).Exclude(exclude).Page(page).PerPage(perPage).Execute()
 	if err != nil {
 		return "", err
 	}
-	for _, p := range projects {
-		if p.Organization.ID == userOrg {
-			return p.ID, nil
+	for _, p := range projects.GetProjects() {
+		if p.AdditionalProperties["organization"] == userOrg {
```
ctreatma: `p.AdditionalProperties["organization"]` will be an object/map;  you'll probably need to do something more like:
```
organization := p.AdditionalProperties["organization"].(map[string]interface{})
organization_id := fmt.Sprintf("%v", organization["id"])
```

internal/init/init.go
```diff
@@ -132,18 +135,19 @@ func (c *Client) NewCommand() *cobra.Command {
...
 }
 
-func getFirstProjectID(s packngo.ProjectService, userOrg string) (string, error) {
-	listOpts := &packngo.ListOptions{}
-	listOpts.Including("organization")
-	listOpts.Excluding("devices", "members", "memberships", "invitations", "max_devices", "ssh_keys", "volumes", "backend_transfer_enabled", "updated_at", "customdata", "event_alert_configuration")
+func getFirstProjectID(s metal.ProjectsApiService, userOrg string) (string, error) {
+	include := []string{"organization"} // []string | Nested attributes to include. Included objects will return their full attributes. Attribute names can be dotted (up to 3 levels) to included deeply nested objects. (optional)
+	exclude := []string{"devices", "members", "memberships", "invitations", "ssh_keys", "volumes", "backend_transfer_enabled", "updated_at", "customdata", "event_alert_configuration"}
+	page := int32(56)    // int32 | Page to return (optional) (default to 1)
```
ctreatma: We should start from page 1

internal/init/init.go
```diff
@@ -132,18 +135,19 @@ func (c *Client) NewCommand() *cobra.Command {
...
+	perPage := int32(56) // int32 | Items returned per page (optional) (default to 10)
```
ctreatma, line 144: Overriding `perPage` is OK, but we'll still need to handle the pagination.  Here's an example of collecting paginated devices in the sweeper action; we'll need something like this for collecting projects: https://github.com/equinix-labs/metal-sweeper-action/blob/main/main.go#L79-L98

internal/init/init.go
```diff
@@ -132,18 +135,19 @@ func (c *Client) NewCommand() *cobra.Command {
...
+	projects, _, err := s.FindProjects(context.Background()).Include(include).Exclude(exclude).Page(page).PerPage(perPage).Execute()
 	if err != nil {
 		return "", err
 	}
-	for _, p := range projects {
-		if p.Organization.ID == userOrg {
-			return p.ID, nil
+	for _, p := range projects.GetProjects() {
+		if p.AdditionalProperties["organization"] == userOrg {
+			return fmt.Sprintf("%v", p.AdditionalProperties["organization"]), nil
```
ctreatma: This function return `project.GetId()` rather than the organization ID

internal/init/init.go
```diff
@@ -85,25 +86,26 @@ func (c *Client) NewCommand() *cobra.Command {
...
+			c.ProjectService = *metalGoClient.ProjectsApi
 
-			user, _, err := c.UserService.Current()
+			user, _, err := c.UserService.FindCurrentUser(context.Background()).Execute()
 			if err != nil {
 				return err
 			}
-			organization := user.DefaultOrganizationID
+			organization := user.AdditionalProperties["default_organization_id"]
 			project := ""
```
ctreatma: We should initialize `project` with `fmt.Sprintf("%v", user.AdditionalProperties["default_project_id"])` so that we use the default if they have one; we should only run `getFirstProjectId` when the user _doesn't_ already have a default.

internal/init/init.go
codinja1188: updated

internal/init/init.go
codinja1188: updated

internal/init/init.go
codinja1188: updated

PR #275, opened 2023-06-06
Updated metal-go client for OS, Metros, Plans

codinja1188:

This PR replaces packngo with metal-go for all interactions with the Equinix Metal API specifically for OS, Metros, Plans.

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -259,7 +259,7 @@ func (c *Client) NewCommand() *cobra.Command {
 	rootCmd.PersistentFlags().StringVar(&c.sortBy, "sort-by", "", "Sort fields for use in 'get' actions. Sort is not supported by all resources.")
 	rootCmd.PersistentFlags().StringVar(&c.sortDir, "sort-dir", "", "Sort field direction for use in 'get' actions. Sort is not supported by all resources.")
 
-	rootCmd.Version = getMetalVersion()
+	rootCmd.Version = c.Version
 	c.rootCmd = rootCmd
 	return c.rootCmd
 }
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -259,7 +259,7 @@ func (c *Client) NewCommand() *cobra.Command {
 	rootCmd.PersistentFlags().StringVar(&c.sortBy, "sort-by", "", "Sort fields for use in 'get' actions. Sort is not supported by all resources.")
 	rootCmd.PersistentFlags().StringVar(&c.sortDir, "sort-dir", "", "Sort field direction for use in 'get' actions. Sort is not supported by all resources.")
 
-	rootCmd.Version = c.Version
+	rootCmd.Version = getMetalVersion()
 	c.rootCmd = rootCmd
 	return c.rootCmd
 }
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -96,7 +96,6 @@ func (c *Client) NewCommand() *cobra.Command {
 			}
 			organization := user.AdditionalProperties["default_organization_id"]
 			project := fmt.Sprintf("%v", user.AdditionalProperties["default_project_id"])
-
 			fmt.Printf("Organization ID [%s]: ", organization)
 
 			defaultOrganizationId := fmt.Sprintf("%v", organization)
@@ -122,7 +121,7 @@ func (c *Client) NewCommand() *cobra.Command {
 			userProj := ""
 			fmt.Scanln(&userProj)
 			if userProj == "" {
-				userProj = fmt.Sprintf("value: %v", organization)
+				userProj = project
 			}
 
 			b, err = formatConfig(userProj, userOrg, token)
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -96,6 +96,7 @@ func (c *Client) NewCommand() *cobra.Command {
 			}
 			organization := user.AdditionalProperties["default_organization_id"]
 			project := fmt.Sprintf("%v", user.AdditionalProperties["default_project_id"])
+
 			fmt.Printf("Organization ID [%s]: ", organization)
 
 			defaultOrganizationId := fmt.Sprintf("%v", organization)
@@ -121,7 +122,7 @@ func (c *Client) NewCommand() *cobra.Command {
 			userProj := ""
 			fmt.Scanln(&userProj)
 			if userProj == "" {
-				userProj = project
+				userProj = fmt.Sprintf("value: %v", organization)
 			}
 
 			b, err = formatConfig(userProj, userOrg, token)
--- a/internal/metros/retrieve.go
+++ b/internal/metros/retrieve.go
@@ -21,6 +21,7 @@
 package metros
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/spf13/cobra"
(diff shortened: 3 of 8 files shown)


internal/init/init.go
```diff
@@ -132,24 +124,6 @@ func (c *Client) NewCommand() *cobra.Command {
 	return initCmd
 }
 
-func getFirstProjectID(s packngo.ProjectService, userOrg string) (string, error) {
```
displague: I think we could preserve this behavior of choosing the first (sometimes only) organization project as the default for metal-cli when none is defined in the user preferences. I believe `default_project_id` preference is set in the console and not all users will have on defined.

internal/cli/root.go
```diff
@@ -219,7 +264,7 @@ func (c *Client) NewCommand() *cobra.Command {
 	rootCmd.PersistentFlags().StringVar(&c.sortBy, "sort-by", "", "Sort fields for use in 'get' actions. Sort is not supported by all resources.")
 	rootCmd.PersistentFlags().StringVar(&c.sortDir, "sort-dir", "", "Sort field direction for use in 'get' actions. Sort is not supported by all resources.")
 
-	rootCmd.Version = c.Version
```
displague: The version is compiled into the metal-cli binary. I don't think we need to change this.

internal/cli/root.go
```diff
@@ -93,6 +103,15 @@ func (c *Client) apiConnect(httpClient *http.Client) error {
 	return nil
 }
 
+func (c *Client) metalApiConnect(httpClient *http.Client) error {
+	configuration := metal.NewConfiguration()
+	configuration.Debug = checkEnvForDebug()
+	configuration.AddDefaultHeader("X-Auth-Token", c.Token())
+	metalgoClient := metal.NewAPIClient(configuration)
```
displague: We'll want to put the UserAgent here, as you did in #276 
https://github.com/equinix/metal-cli/pull/276/files#diff-aa84670cfbda5699ea29a1dad10df2df47f189c2748c57f6ea3efff057313658R110.

internal/cli/root.go
```diff
@@ -169,6 +188,32 @@ func (c *Client) API(cmd *cobra.Command) *packngo.Client {
...
+
+		err := c.metalApiConnect(httpClient)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+	return c.metalApiClient
+}
+
+func getMetalVersion() string {
```
displague: If we stick to c.Version for rootCmd.Version below, I think we aren't using this function and can remove it.

displague (changes requested):

displague:
Something that is not apparent in the PRs, that we'll need to be aware of with the metal-go conversions, is that the `-o yaml` and `-o json` formats will be different based on packngo or metal-go type definitions. 

Generally, we should be getting more fields from the `metal-go` exports, but we may experience some renamed fields.  Users that depend on `metal get ... -o json` format will need to be aware of this in the release notes. (https://github.com/TritonDataCenter/triton/issues/325 for example)

internal/plans/retrieve.go
```diff
@@ -37,15 +38,15 @@ func (c *Client) Retrieve() *cobra.Command {
...
-				return fmt.Errorf("Could not list Plans: %w", err)
+				return fmt.Errorf("could not list plans: %w", err)
 			}
 
-			data := make([][]string, len(plans))
+			data := make([][]string, len(plans.GetPlans()))
 
-			for i, p := range plans {
-				data[i] = []string{p.ID, p.Slug, p.Name}
+			for i, p := range plans.GetPlans() {
```
displague: Let's use `.GetPlans()` on [line 53](https://github.com/equinix/metal-cli/pull/275/files#diff-819816e78ca64595a797aab7da73730c936b25cce73b5c995770fcad7913e88dR53) too:

```
return c.Out.Output(plans.GetPlans(), header, &data)
```

Same for [Metros](https://github.com/equinix/metal-cli/pull/275/files#diff-d7dcdf7eb8b913a7e9d1b0b3d933682a8d8f28b189184fed038ec61662e918b4R53) and [OS](https://github.com/equinix/metal-cli/pull/275/files#diff-d1256b9e6f104a409c0395ad7295d0ec047735d5dccfbcabeedf4a99986a7cb9R58).

This preserves (as much as possible) the `-o yaml` and `-o json` format functionality.

internal/init/init.go
ctreatma: This was covered in #274

internal/cli/root.go
ctreatma: Also covered in #274

internal/plans/retrieve.go
codinja1188: updated

internal/cli/root.go
codinja1188: removed

internal/metros/retrieve.go
```diff
@@ -38,18 +39,18 @@ func (c *Client) Retrieve() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			metros, _, err := c.Service.List(c.Servicer.ListOptions(nil, nil))
+			metros, _, err := c.Service.FindMetros(context.Background()).Execute()
```
ctreatma: nit: you could change `metros` to `metrosList` here and then define `metros := metrosList.GetMetros()` after line 45 to reduce duplicated calls to `GetMetros()`.

internal/os/retrieve.go
```diff
@@ -38,23 +39,23 @@ func (c *Client) Retrieve() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			oss, _, err := c.Service.List()
+			oss, _, err := c.Service.FindOperatingSystems(context.Background()).Execute()
```
ctreatma: nit: same as with metros; you could rename `oss` to `ossList` here, and then have a single `oss := ossList.GetOperatingSystems()` rather than having multiple calls to `GetOperatingSystems()`

internal/plans/retrieve.go
```diff
@@ -37,19 +38,19 @@ func (c *Client) Retrieve() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			plans, _, err := c.Service.List(c.Servicer.ListOptions(nil, nil))
+			plans, _, err := c.Service.FindPlans(context.Background()).Execute()
```
ctreatma: Same comment here as in other subcommands

internal/metros/retrieve.go
codinja1188: updated

internal/os/retrieve.go
codinja1188: updated

internal/plans/retrieve.go
codinja1188: updated

ctreatma (approved):
Looks good to me!  I built and ran each of these subcommands locally and compared the JSON output to the latest release of metal-go, and the only differences I saw were additional fields from the metal-go-based commands.

PR #276, opened 2023-06-06
Updated metal-go client for projects sub-commands

codinja1188:

This PR replaces packngo with metal-go for all interactions with the Equinix Metal API specifically for projects sub-commands

### DIFF
--- a/internal/projects/retrieve.go
+++ b/internal/projects/retrieve.go
@@ -21,10 +21,11 @@
 package projects
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/equinix/metal-cli/internal/outputs"
-	"github.com/packethost/packngo"
+	pager "github.com/equinix/metal-cli/internal/pagination"
 	"github.com/spf13/cobra"
 )
 
@@ -50,8 +51,8 @@ func (c *Client) Retrieve() *cobra.Command {
 			if projectID != "" && projectName != "" {
 				return fmt.Errorf("Must specify only one of project-id and project name")
 			}
-
 			inc := []string{}
+			exc := []string{}
 
 			// only fetch extra details when rendered
 			switch c.Servicer.Format() {
@@ -59,10 +60,8 @@ func (c *Client) Retrieve() *cobra.Command {
 				inc = append(inc, "members")
 			}
 
-			listOpts := c.Servicer.ListOptions(inc, nil)
-
 			if projectID == "" {
-				projects, _, err := c.ProjectService.List(listOpts)
+				projects, err := pager.GetAllProjects(c.ProjectService, inc, exc)
 				if err != nil {
 					return fmt.Errorf("Could not list Projects: %w", err)
 				}
@@ -71,13 +70,13 @@ func (c *Client) Retrieve() *cobra.Command {
 				if projectName == "" {
 					data = make([][]string, len(projects))
 					for i, p := range projects {
-						data[i] = []string{p.ID, p.Name, p.Created}
+						data[i] = []string{p.GetId(), p.GetName(), p.GetCreatedAt().String()}
 					}
 				} else {
 					data = make([][]string, 0)
 					for _, p := range projects {
-						if p.Name == projectName {
-							data = append(data, []string{p.ID, p.Name, p.Created})
+						if p.GetName() == projectName {
+							data = append(data, []string{p.GetId(), p.GetName(), p.GetCreatedAt().String()})
 							break
 						}
 					}
@@ -89,15 +88,14 @@ func (c *Client) Retrieve() *cobra.Command {
 				header := []string{"ID", "Name", "Created"}
 				return c.Out.Output(projects, header, &data)
 			} else {
-				getOpts := &packngo.GetOptions{Includes: listOpts.Includes, Excludes: listOpts.Excludes}
-				p, _, err := c.ProjectService.Get(projectID, getOpts)
+				p, _, err := c.ProjectService.FindProjectById(context.Background(), projectID).Include(inc).Exclude(exc).Execute()
 				if err != nil {
 					return fmt.Errorf("Could not get Project: %w", err)
 				}
 
 				data := make([][]string, 1)
 
-				data[0] = []string{p.ID, p.Name, p.Created}
+				data[0] = []string{p.GetId(), p.GetName(), p.GetCreatedAt().String()}
 				header := []string{"ID", "Name", "Created"}
 				return c.Out.Output(p, header, &data)
 			}
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -29,6 +29,7 @@ import (
 	"syscall"
 
 	metal "github.com/equinix-labs/metal-go/metal/v1"
+	pager "github.com/equinix/metal-cli/internal/pagination"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
 	"sigs.k8s.io/yaml"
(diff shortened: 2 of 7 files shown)


internal/projects/bgpsessions.go
```diff
@@ -41,19 +41,17 @@ func (c *Client) BGPSessions() *cobra.Command {
   metal project bgp-sessions --project-id 50693ba9-e4e4-4d8a-9eb2-4840b11e9375`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			listOpt := c.Servicer.ListOptions(nil, nil)
-			getOpts := &packngo.GetOptions{Includes: listOpt.Includes, Excludes: listOpt.Excludes}
-			p, _, err := c.BGPConfigService.Get(projectID, getOpts)
+			p, _, err := c.BGPConfigService.FindProjectBgpSessions(context.Background(), projectID).Execute()
```
ctreatma: Judging from `packngo`, BGP sessions are paginated: https://github.com/packethost/packngo/blob/master/projects.go#L154-L175

The spec doesn't say it's paginated, so we'll have to address that in order to handle pagination for BGP settings: https://deploy.equinix.com/developers/api/metal#tag/BGP/operation/findProjectBgpSessions

internal/projects/retrieve.go
```diff
@@ -51,33 +50,23 @@ func (c *Client) Retrieve() *cobra.Command {
...
-			switch c.Servicer.Format() {
-			case outputs.FormatJSON, outputs.FormatYAML:
-				inc = append(inc, "members")
-			}
-
-			listOpts := c.Servicer.ListOptions(inc, nil)
-
 			if projectID == "" {
-				projects, _, err := c.ProjectService.List(listOpts)
+				projects, _, err := c.ProjectService.FindProjects(context.Background()).Execute()
```
ctreatma: - The `listOpts` need to be converted to `Includes`, otherwise we're not making the same API calls
- You'll need to use the `getAllProjects` function that was added for the `init` command in order to handle pagination

internal/projects/bgpsessions.go
ctreatma: I looked at the API code and it's possible that in this case, the spec is right and packngo is wrong; the code for rendering a list of BGP sessions does not include pagination details.

internal/projects/retrieve.go
codinja1188: updated as per comment

internal/projects/bgpsessions.go
codinja1188: Then, Is it fine to close conversation?

ctreatma: Yeah, this is not an issue after all

internal/projects/retrieve.go
```diff
@@ -89,15 +88,14 @@ func (c *Client) Retrieve() *cobra.Command {
 				header := []string{"ID", "Name", "Created"}
 				return c.Out.Output(projects, header, &data)
 			} else {
-				getOpts := &packngo.GetOptions{Includes: listOpts.Includes, Excludes: listOpts.Excludes}
-				p, _, err := c.ProjectService.Get(projectID, getOpts)
+				p, _, err := c.ProjectService.FindProjectById(context.Background(), projectID).Execute()
```
ctreatma: The includes/excludes have been lost here

codinja1188: updated

internal/facilities/retrieve.go
```diff
@@ -38,22 +39,22 @@ func (c *Client) Retrieve() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			facilities, _, err := c.Service.List(c.Servicer.ListOptions(nil, nil))
+			facilities, _, err := c.Service.FindFacilities(context.Background()).Execute()
```
ctreatma: If you rename `facilities` to `facilitiesList` here, you can do `facilities := facilitiesList.GetFacilities()` and simplify some of the changes later in the function.

internal/facilities/retrieve.go
```diff
@@ -38,22 +39,22 @@ func (c *Client) Retrieve() *cobra.Command {
...
 			}
-			data := make([][]string, len(facilities))
+			data := make([][]string, len(facilities.GetFacilities()))
 
-			for i, facility := range facilities {
+			for i, facility := range facilities.GetFacilities() {
 				var metro string
 				if facility.Metro != nil {
-					metro = facility.Metro.Code
+					metro = *facility.GetMetro().Code
```
ctreatma:
```suggestion
					metro = facility.Metro.GetCode()
```

internal/projects/bgpconfig.go
```diff
@@ -40,16 +40,14 @@ func (c *Client) BGPConfig() *cobra.Command {
   metal project bgp-config --project-id 50693ba9-e4e4-4d8a-9eb2-4840b11e9375`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			listOpt := c.Servicer.ListOptions(nil, nil)
-			getOpts := &packngo.GetOptions{Includes: listOpt.Includes, Excludes: listOpt.Excludes}
-			p, _, err := c.BGPConfigService.Get(projectID, getOpts)
+			p, _, err := c.BGPConfigService.FindBgpConfigByProject(context.Background(), projectID).Execute()
```
ctreatma: it looks like includes/excludes have been lost here as well?

internal/projects/bgpsessions.go
ctreatma: it looks like this has also lost includes/excludes

codinja1188: Undefined Included and excludes for this API

internal/projects/bgpconfig.go
codinja1188: Updated but empty

internal/facilities/retrieve.go
codinja1188: updated

internal/facilities/retrieve.go
codinja1188: updated

internal/projects/bgpsessions.go
ctreatma: @vasubabu it looks like the other includes/excludes were restored, but this one is still missing

codinja1188: > @vasubabu it looks like the other includes/excludes were restored, but this one is still missing
This API doesn't support Includes/Excludes. https://github.com/equinix-labs/metal-go/blob/main/docs/BGPApi.md#FindProjectBgpSessions

internal/facilities/facility.go
```diff
@@ -21,14 +21,14 @@
...
+	metal "github.com/equinix-labs/metal-go/metal/v1"
 	"github.com/equinix/metal-cli/internal/outputs"
-	"github.com/packethost/packngo"
 	"github.com/spf13/cobra"
 )
 
 type Client struct {
 	Servicer Servicer
-	Service  packngo.FacilityService
+	Service  metal.FacilitiesApiService
```
ctreatma: One thing we lose with this migration is the automatic output of `Deprecation` headers.  `packngo` prints a warning to stdout when an API response includes a `Deprecation` header; while I'm not a fan of the existing behavior, we still probably need to replicate it in `metal-cli` as part of the `metal-go` transition.  Can you break the `facilities` subcommands into a separate PR?

codinja1188: sure, will do it

ctreatma (approved):
There are some differences in JSON output (packngo includes some empty fields where metal-go omits them, and metal-go sometimes has fields that packngo doesn't know about), but overall they should be sufficiently similar.

PR #284, opened 2023-06-16
Updated metal-go client for sub-command orgs

codinja1188:

This PR replaces packngo with metal-go for all interactions with the Equinix Metal API specifically for orgs sub commands

DISCUSSION POINTS:

### DIFF
--- a/internal/organizations/update.go
+++ b/internal/organizations/update.go
@@ -21,9 +21,10 @@
 package organizations
 
 import (
+	"context"
 	"fmt"
 
-	"github.com/packethost/packngo"
+	metal "github.com/equinix-labs/metal-go/metal/v1"
 	"github.com/spf13/cobra"
 )
 
@@ -39,7 +40,7 @@ func (c *Client) Update() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			req := &packngo.OrganizationUpdateRequest{}
+			req := metal.NewOrganizationInput()
 
 			if name != "" {
 				req.Name = &name
@@ -57,14 +58,18 @@ func (c *Client) Update() *cobra.Command {
 				req.Logo = &logo
 			}
 
-			org, _, err := c.Service.Update(organizationID, req)
+			if website != "" {
+				req.Website = &website
+			}
+
+			org, _, err := c.Service.UpdateOrganization(context.Background(), organizationID).OrganizationInput(*req).Execute()
 			if err != nil {
 				return fmt.Errorf("Could not update Organization: %w", err)
 			}
 
 			data := make([][]string, 1)
 
-			data[0] = []string{org.ID, org.Name, org.Created}
+			data[0] = []string{org.GetId(), org.GetName(), org.GetCreatedAt().String()}
 			header := []string{"ID", "Name", "Created"}
 
 			return c.Out.Output(org, header, &data)
--- a/test/e2e/organization_test.go
+++ b/test/e2e/organization_test.go
@@ -1,0 +1,51 @@
+package e2e
+
+import (
+	"testing"
+
+	root "github.com/equinix/metal-cli/internal/cli"
+	"github.com/equinix/metal-cli/internal/organizations"
+	outputPkg "github.com/equinix/metal-cli/internal/outputs"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCli_Organization(t *testing.T) {
+	subCommand := "organization"
+	consumerToken := ""
+	apiURL := ""
+	Version := "metal"
+	rootClient := root.NewClient(consumerToken, apiURL, Version)
+	type fields struct {
+		MainCmd  *cobra.Command
+		Outputer outputPkg.Outputer
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    *cobra.Command
+		cmdFunc func(*testing.T, *cobra.Command)
+	}{
+		{
+			name: "get",
+			fields: fields{
+				MainCmd:  organizations.NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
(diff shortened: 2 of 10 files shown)


internal/organizations/update.go
```diff
@@ -54,17 +57,28 @@ func (c *Client) Update() *cobra.Command {
 			}
 
 			if logo != "" {
-				req.Logo = &logo
+				logoData, err := os.OpenFile(logo, os.O_RDWR, 0755)
```
ctreatma: This is new logic that didn't exist anywhere in packngo, and I don't think it's the approach we want to take.  The spec says that the `logo` for an `OrganizationInput` (i.e., when creating or updating an organization) but that may be wrong.  We need confirmation from the API team before we can pick a path forward.

displague: I updated my personal org logo in the console with the network inspector on.  In this scenario, the logo field is submitted as a [base64 encoded Data URL](https://developer.mozilla.org/en-US/docs/web/http/basics_of_http/data_urls).

```
"logo": "data:image/jpeg;base64,/9j...RRRF//Z"
```

displague: I also verified that PNG was permitted. I don't know what other formats (including URL formats, such as https), if any, are permitted.

ctreatma: > PNG was permitted

Do you mean that you submitted a PNG as a data URI, or that you submitted the raw contents of a PNG, or both?  In either of these cases, what was the value of the `logo` attribute in the response?  Did the logo render correctly in the portal in both cases?

displague: For UserData, we introduced a `--userdata-file` argument as a complement to `--userdata`. Perhaps we could follow that pattern with a `--logo-file` argument.

displague: I also tried `data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg"/>`, but this was rejected. Converting the contents to base64 and submitting it in that format was accepted.

Upon submission, the `logo` value is mutated. The result is an S3 archive URL. `logo_thumb` is also returned in the response, referring to another S3 housed image.

ctreatma: From API team:  `logo` is a string containing base64-encoded image data, so the `packngo` behavior was correct, but the documentation that logo is a URL is incorrect.  The spec has been updated but is not deployed yet; once it's out we can regenerate `metal-go`, cut a new release, update `metal-cli` to use that release, and then proceed with this PR.  I don't think we should add the `--logo-file` argument in the scope of this PR.

cprivitere:
Needs a rebase due to conflicts with pager.go, sadly. :(

internal/organizations/update.go
ctreatma: Upgrading to [`metal-go@v0.11.0`](https://github.com/equinix-labs/metal-go/releases/tag/v0.11.0) will allow you to pass in the `logo` as a string, similar to the old `packngo`-based code.  I opened #295 to add a logo-file option; IMO we should do that in a separate PR.

internal/organizations/organization_test.go
```diff
@@ -0,0 +1,176 @@
...
+			fields: fields{
+				MainCmd:  NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				t.Helper()
+				root := c.Root()
+				out := &bytes.Buffer{}
+				root.SetArgs([]string{subCommand, "update", "-i", "3bd5bf07-6094-48ad-bd03-d94e8712fdc8", "--name", "test-org2"})
```
ctreatma: It would be nice if we could structure these organization tests a little differently so that we could create a test organization, capture the ID of that organization, and then pass that ID in to subsequent update/payment/delete tests.

internal/organizations/organization_test.go
```diff
@@ -0,0 +1,176 @@
...
+			fields: fields{
+				MainCmd:  NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				t.Helper()
+				root := c.Root()
+				out := &bytes.Buffer{}
+				root.SetArgs([]string{subCommand, "create", "-n", "test-org1", "-w", "www.metal.equinix.com"})
```
ctreatma: In the terraform provider acceptance tests, we use a naming convention (`tfacc-<resource>-<suffix>`, e.g. `tfacc-org-asdlfkjlkj`) for all resources that are created by the tests so that we can easily identify those resources.  We should adopt a similar naming convention in this project, something like `cli-acc-<resource>-<suffix>`.

End-to-end tests should be written in such a way that resources created by the tests are cleaned up, but using a naming convention allows us to easily identify test resources if we need to clean them up separately for some reason.

```suggestion
				root.SetArgs([]string{subCommand, "create", "-n", "cli-acc-org-test1", "-w", "www.metal.equinix.com"})
```

codinja1188: updated

internal/organizations/organization_test.go
codinja1188: updated

test/e2e/organization_test.go
```diff
@@ -0,0 +1,142 @@
...
+					if err != nil {
+						t.Errorf("Error setting environment variable: %v ", err)
+					}
+				} else {
+					t.Errorf("No ID found")
+				}
+			},
+		},
+		{
+			name: "update",
```
ctreatma: Since org creation doesn't work in metal-cli at the moment, let's remove these `update` and `payment-method` tests so that we're only testing `get`.  We can come back and fill in tests later as part of fixing org creation.

codinja1188: updated

ctreatma (approved):

PR #287, opened 2023-06-19
Updated metal-go client for sub-command events

codinja1188:

This PR replaces packngo with metal-go for all interactions with the Equinix Metal API specifically for Events sub commands

DISCUSSION POINTS:

### DIFF
--- a/internal/events/retrieve.go
+++ b/internal/events/retrieve.go
@@ -21,10 +21,12 @@
 package events
 
 import (
+	"context"
 	"fmt"
 
+	metal "github.com/equinix-labs/metal-go/metal/v1"
 	"github.com/equinix/metal-cli/internal/outputs"
-	"github.com/packethost/packngo"
+	pager "github.com/equinix/metal-cli/internal/pagination"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


ctreatma:
This overlaps with #276 and should be rebased once that PR is merged

cprivitere:
Seems to be working!

internal/events/retrieve.go
```diff
@@ -53,47 +55,47 @@ func (c *Client) Retrieve() *cobra.Command {
...
 			inc := []string{}
+			exc := []string{}
 
 			// only fetch extra details when rendered
 			switch c.Servicer.Format() {
 			case outputs.FormatJSON, outputs.FormatYAML:
 				inc = append(inc, "relationship")
 			}
 
-			listOpt := c.Servicer.ListOptions(inc, nil)
```
ctreatma, line 67: The `packngo` code called `c.Servicer.ListOptions(inc, nil)`, which would populate the includes and excludes based on the `--include` and `--exclude` command-line flags.

It looks like that was lost in the `metal-go` conversion. @vasubabu could you open a follow-up PR to hook this subcommand up to the `c.Servicer.Include()` and `c.Servicer.Exclude()` functions that you added?

PR #300, opened 2023-07-07
Added Unit test case for metro sub command

codinja1188:


internal/metros/metro_test.go
```diff
@@ -0,0 +1,82 @@
...
+				if !strings.Contains(out.String(), "Get information on Metro locations") {
+					t.Error("expected output should include list all known hardware")
+				}
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rootCmd := rootClient.NewCommand()
+			rootCmd.AddCommand(tt.fields.MainCmd)
```
ctreatma:
```suggestion
			rootCmd.AddCommand(tt.fields.MainCmd)
			tt.cmdFunc(t, tt.fields.MainCmd)
```

codinja1188: updated

internal/metros/metro_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"strings"
+	"testing"
+
+	root "github.com/equinix/metal-cli/internal/cli"
+	outputPkg "github.com/equinix/metal-cli/internal/outputs"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCli_Metros(t *testing.T) {
```
ctreatma: This branch needs to be rebased on the latest from `main` and this test has to be skipped in short mode.

codinja1188: rebased

internal/metros/metro_test.go
```diff
@@ -0,0 +1,56 @@
...
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				out := &bytes.Buffer{}
+				root.SetArgs([]string{subCommand, "get"})
+				root.SetOutput(out)
+				err := root.Execute()
+				assert.NoError(t, err)
```
ctreatma: For the basic `get` tests (in this PR and in others), we should have additional assertions to validate that  we got the expected results.  We don't need to check the returned metro data exhaustively, but it would be good to spot check that we got at least one metro, and/or that the response contains the expected field names.

ctreatma:

internal/metros/metro_test.go
codinja1188: updated

test/e2e/metro_test.go
```diff
@@ -0,0 +1,69 @@
...
+				os.Stdout = rescueStdout
+				if !strings.Contains(string(out[:]), "Seattle") &&
+					!strings.Contains(string(out[:]), "Tokyo") &&
+					!strings.Contains(string(out[:]), "Sydney") &&
+					!strings.Contains(string(out[:]), "Toronto") &&
+					!strings.Contains(string(out[:]), "Chicago") &&
+					!strings.Contains(string(out[:]), "Dallas") {
+					t.Error("expected output should include Seattle, Tokyo, Sydney, Toronto, Chicago and Dallas by NAME")
+				}
+			},
```
ctreatma, line 58: Does the `metro` subcommand support filtering (e.g., by `code`)?  If so it would be good to add a test for that.

codinja1188: The metro subcommand doesn't support  filtering

ctreatma (approved):

PR #304, opened 2023-07-12
Added Unit test case for plans sub commands

codinja1188:

This PR was added to unit test cases for plans subcommands.

### DIFF
--- a/test/e2e/plan_test.go
+++ b/test/e2e/plan_test.go
@@ -1,0 +1,123 @@
+package e2e
+
+import (
+	"io"
+	"os"
+	"strings"
+	"testing"
+
+	root "github.com/equinix/metal-cli/internal/cli"
+	outputPkg "github.com/equinix/metal-cli/internal/outputs"
+	"github.com/equinix/metal-cli/internal/plans"
+	"github.com/spf13/cobra"
+)
+
+func TestCli_Plans(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
+	subCommand := "plans"
+	consumerToken := ""
+	apiURL := ""
+	Version := "metal"
+	rootClient := root.NewClient(consumerToken, apiURL, Version)
+	type fields struct {
+		MainCmd  *cobra.Command
+		Outputer outputPkg.Outputer
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    *cobra.Command
+		cmdFunc func(*testing.T, *cobra.Command)
+	}{
+		{
+			name: "get",
+			fields: fields{
+				MainCmd:  plans.NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				root.SetArgs([]string{subCommand, "get"})
+				rescueStdout := os.Stdout
+				r, w, _ := os.Pipe()
+				os.Stdout = w
+				if err := root.Execute(); err != nil {
+					t.Error(err)
+				}
+				w.Close()
+				out, _ := io.ReadAll(r)
+				os.Stdout = rescueStdout
+				if !strings.Contains(string(out[:]), "m3.small.x86") &&
+					!strings.Contains(string(out[:]), "m3.large.x86") &&
+					!strings.Contains(string(out[:]), "c3.medium.x86") &&
+					!strings.Contains(string(out[:]), "c3.small.x86") &&
+					!strings.Contains(string(out[:]), "x2.xlarge.x86") &&
+					!strings.Contains(string(out[:]), "x3.xlarge.x86") {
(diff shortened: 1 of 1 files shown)


internal/plans/plan_test.go
```diff
@@ -0,0 +1,56 @@
...
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				out := &bytes.Buffer{}
+				root.SetArgs([]string{subCommand, "get"})
+				root.SetOutput(out)
+				err := root.Execute()
+				assert.NoError(t, err)
```
ctreatma: I think the `plans` subcommand warrants more detailed testing.  For example:  it would be useful to also have a test that uses relevant command line flags and confirms that the response has the expected data (e.g., if I specify a filter `slug=m3.small.x86`, I get back plan data that has a `slug` of `m3.small.x86`).

ctreatma:

ctreatma:

internal/plans/plan_test.go
```diff
@@ -0,0 +1,108 @@
...
+			fields: fields{
+				MainCmd:  NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				out := &bytes.Buffer{}
+				root.SetArgs([]string{subCommand, "get", "--filter ", "type=standard"})
+				root.SetOutput(out)
```
ctreatma: For some reason our CLI doesn't use this for output.  Instead, output is written to the `Outputer` that was configured when the command was instantiated:
- Call to `c.Out.Output` in command: https://github.com/equinix/metal-cli/blob/09eca82c9f66c8a3b8c825f6caa9669de05dc6bf/internal/plans/retrieve.go#L66
- Line in this test where the `Outputer` is passed in: https://github.com/equinix/metal-cli/pull/304/files#diff-b14af556f6d82ab43fd002b0468b08393314c95411f5b84e90dc9a2523716fa1R79

test/e2e/plan_test.go
```diff
@@ -0,0 +1,109 @@
...
+					t.Error("expected output should include m3.small.x86, m3.large.x86, c3.medium.x86 and c3.small.x86 by SLUG")
+				}
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rootCmd := rootClient.NewCommand()
+			rootCmd.AddCommand(tt.fields.MainCmd)
```
ctreatma, line 119: One thing you could try is getting rid of `tt.fields.MainCmd`, and instead doing:

```go
out := &bytes.Buffer{}
testCommand := NewClient(rootClient, outputPkg.Outputer(out)).NewCommand()
rootCmd.AddCommand(testCommand)
tt.cmdFunc(t, rootCmd, out)
```
I haven't tried that, though, so I don't know if `outputPkg.Outputer` will accept a byte buffer.

ctreatma (dismissed):
testing

internal/plans/plan_test.go
codinja1188: updated

test/e2e/plan_test.go
```diff
@@ -0,0 +1,123 @@
...
+					!strings.Contains(string(out[:]), "c3.medium.x86") &&
+					!strings.Contains(string(out[:]), "c3.small.x86") &&
+					!strings.Contains(string(out[:]), "x2.xlarge.x86") &&
+					!strings.Contains(string(out[:]), "x3.xlarge.x86") {
+					t.Error("expected output should include m3.small.x86, m3.large.x8, c3.medium.x86, c3.small.x86, x2.xlarge.x86 and x3.xlarge.x86 by SLUG")
+				}
+			},
+		},
+		{
+			name: "get",
```
ctreatma: Nit: it would be helpful to provide more descriptive/unique names for each of these tests.  This one could be called `get_by_slug`; the one below this one could be `get_by_type`.  More descriptive names make it easier to understand which test code corresponds to which test output.

ctreatma (approved):

test/e2e/plan_test.go
```diff
@@ -0,0 +1,123 @@
...
+		{
+			name: "get_by_type",
+			fields: fields{
+				MainCmd:  plans.NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				root.SetArgs([]string{subCommand, "get", "--token", os.Getenv("METAL_AUTH_TOKEN"), "--filter", "type=standard"})
```
ctreatma, line 96: @vasubabu note that #309 was merged; that PR fixed environment lookup for testing, so if you rebase your remaining open PRs on the latest from `main` you won't need the `--token` flag any more in tests.

PR #305, opened 2023-07-12
Added Unit test case for events sub commands

codinja1188:
Added Unit test case for events sub commands

### DIFF
--- a/test/e2e/events/deviceeventstest/device_event_test.go
+++ b/test/e2e/events/deviceeventstest/device_event_test.go
@@ -1,0 +1,94 @@
+package eventsprojtest
+
+import (
+	"io"
+	"os"
+	"strings"
+	"testing"
+
+	root "github.com/equinix/metal-cli/internal/cli"
+	"github.com/equinix/metal-cli/internal/events"
+	outputPkg "github.com/equinix/metal-cli/internal/outputs"
+	"github.com/equinix/metal-cli/test/helper"
+	"github.com/spf13/cobra"
+)
+
+func TestCli_Events_Get(t *testing.T) {
+	var projectId, deviceId string
+	var err error
+	var status bool
+	subCommand := "event"
+	consumerToken := ""
+	apiURL := ""
+	Version := "metal"
+	rootClient := root.NewClient(consumerToken, apiURL, Version)
+	type fields struct {
+		MainCmd  *cobra.Command
+		Outputer outputPkg.Outputer
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    *cobra.Command
+		cmdFunc func(*testing.T, *cobra.Command)
+	}{
+		{
+			name: "get_events_by_dev_id",
+			fields: fields{
+				MainCmd:  events.NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+
+				projectId, err = helper.CreateTestProject("metal-cli-events-pro")
+				if err != nil {
+					t.Error(err)
+				}
+				deviceId, err = helper.CreateTestDevice(projectId, "metal-cli-events-dev")
+				if err != nil {
+					t.Error(err)
+				}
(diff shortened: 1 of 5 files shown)


internal/events/event_test.go
```diff
@@ -0,0 +1,120 @@
...
+			name: "get",
+			fields: fields{
+				MainCmd:  NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				out := &bytes.Buffer{}
+				root.SetArgs([]string{subCommand, "get", "-p", "1867ee8f-6a11-470a-9505-952d6a324040"})
```
ctreatma: These IDs should, at minimum, be declared as variables, and should probably be read from the environment so that a person running these tests can pass in values that make sense for them.

ctreatma:

test/e2e/event_test.go
```diff
@@ -0,0 +1,148 @@
...
+		{
+			name: "get_by_projID",
+			fields: fields{
+				MainCmd:  events.NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				root.SetArgs([]string{subCommand, "get", "-p", os.Getenv("METAL_PROJ_ID")})
```
ctreatma: In general we should create test resources rather than reading IDs from the environment.  Exceptions to that guideline include organization ID and hardware reservation ID, because those are harder or impossible to create dynamically.

I think a good set of tests for events would:
(1) create a test project (with a standard naming convention so we can easily identify test projects created by metal-cli tests)
(2) create a device in that test project
(3) run various `events` subcommands and validate that the output matches expectations

However: for now, I think it's sufficient to just have the first `get` test and save the more complex test cases for a later PR.

codinja1188: Updated as per review comments.

ctreatma:
Note that the tests are failing due to what looks like a 10-minute timeout (the timeout is happening in the test code, it's not an API request timeout).  https://github.com/equinix/metal-cli/actions/runs/5648492098/job/15300834778

ctreatma:
I think this is failing because of the sheer number of available events; the paginator is trying to load all of them and there are so many that it's taking too long for the test to pass.  For now, lets update the test to do a `metal project get`, capture the ID of the first project, and then run `metal events get -p <first project ID>`.  We should probably only check that the events command succeeded for now, instead of checking for specific event types in the output; we can come back later and improve on that.

test/e2e/event_test.go
```diff
@@ -0,0 +1,109 @@
...
+		{
+			name: "get Project Id",
+			fields: fields{
+				MainCmd:  projects.NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
+				root.SetArgs([]string{"project", "get"})
```
ctreatma: This should create a project instead of getting an existing project so that the test will work the same way in CI as it does locally.

The test helpers you created in a different PR (or PRs?) would be useful here to allow you to bypass `metal-cli` for test setup & teardown.

test/e2e/events/deviceeventstest/device_event_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}{
+		{
+			name: "get_events_by_dev_id",
+			fields: fields{
+				MainCmd:  events.NewClient(rootClient, outputPkg.Outputer(&outputPkg.Standard{})).NewCommand(),
+				Outputer: outputPkg.Outputer(&outputPkg.Standard{}),
+			},
+			want: &cobra.Command{},
+			cmdFunc: func(t *testing.T, c *cobra.Command) {
+				root := c.Root()
```
ctreatma, line 43: You even added a device events test!  Nice!

ctreatma (approved):

PR #306, opened 2023-07-13
ci: split end-to-end tests into a separate workflow

ctreatma:
The end-to-end tests require an API token, which means they can't be run automatically for external contributions.

The end-to-end test workflow is configured to require approval for contributions from an external fork, following the example here: https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v3
+      with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
+    - name: Get dependencies
+      run: go mod download
+    - name: Run end-to-end tests
+      run: go test -v -cover -parallel 4 ./test/e2e
+      env:
+        METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
```
ctreatma: I'm leaning towards removing this for the initial PR; the workflow will fail because the credential isn't configured, but that would enable us to learn how it feels to trigger/approve the workflow first, and then add the secret when we feel more confident in this approach.

codinja1188: Is it needs a previous job to run?

ctreatma: Yes, the approval job has to complete before this job can run.  The approval job uses the `external` environment if the PR comes from an external repo, and that environment is configured to require approval for all workflows, so if no one approves the workflow, the approval job will never finish, and the test job will not run.

codinja1188 (approved):
/lgtm

.github/workflows/e2e-test.yml
ctreatma: I'm making the call to remove the secret from the initial implementation so we can see how the workflow works first.
```suggestion
```

PR #312, opened 2023-07-21
ci: only run end-to-end tests if the code changed

ctreatma:


displague (approved):

.github/workflows/e2e-test.yml
```diff
@@ -9,6 +9,10 @@ name: End-to-end Tests
 
 on:
   pull_request_target:
+    paths:
+    - '**.go'
+    - 'go.mod'
+    - 'go.sum'
   workflow_dispatch:
```
displague, line 16: should `.github/` be included?

ctreatma: The `pull_request_target` trigger runs in the context of the base branch, so changes to the workflow file won't take effect until they're merged.

PR #315, opened 2023-07-27
fix: Updated metal-go client for sub-commands Virtual Network

codinja1188:


internal/vlan/create.go
```diff
@@ -45,28 +46,30 @@ func (c *Client) Create() *cobra.Command {
...
 
-			n, _, err := c.Service.Create(req)
+			n, _, err := c.Service.CreateVirtualNetwork(context.Background(), projectID).VirtualNetworkCreateInput(*virtualNetworkCreateInput).Include(nil).Exclude(nil).Execute()
 			if err != nil {
 				return fmt.Errorf("Could not create ProjectVirtualNetwork: %w", err)
 			}
 
 			data := make([][]string, 1)
 
 			// TODO(displague) metro is not in the response
```
displague, line 69: nit: This TODO can probably be removed now

displague: omitting FacilityCode based on https://github.com/equinix-labs/metal-go/issues/117 and the deprecation of facilities is reasonable.

Additionally, since the introduction of metros, VLAN's are not facility scoped. They are metro scoped. Removing the confusing facility detail makes good sense.

internal/vlan/retrieve.go
```diff
@@ -41,19 +41,28 @@ func (c *Client) Retrieve() *cobra.Command {
...
 
-			data := make([][]string, len(vnets.VirtualNetworks))
-
-			for i, n := range vnets.VirtualNetworks {
-				data[i] = []string{n.ID, n.Description, strconv.Itoa(n.VXLAN), n.FacilityCode, n.CreatedAt}
+			for i, n := range VirtualNetworks {
+				data[i] = []string{n.GetId(), n.GetDescription(), strconv.Itoa(int(n.GetVxlan()))}
 			}
-			header := []string{"ID", "Description", "VXLAN", "Facility", "Created"}
+			header := []string{"ID", "Description", "VXLAN"}
```
displague: We're going to want to expose metro.

internal/vlan/retrieve.go
```diff
@@ -41,19 +41,28 @@ func (c *Client) Retrieve() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			vnets, _, err := c.Service.List(projectID, c.Servicer.ListOptions(nil, nil))
+			request := c.Service.FindVirtualNetworks(cmd.Context(), projectID).Include(nil).Exclude(nil)
+			filters := c.Servicer.Filters()
+			if filters["facility"] != "" {
+				request = request.Facility(filters["facility"])
```
displague, line 47: Keeping the filter by facility functionality, if it still works, can't hurt - I think. Since it is in the API spec / SDK and the code path is already here, I agree with updating the as-is code.

internal/vlan/create.go
```diff
@@ -45,28 +46,30 @@ func (c *Client) Create() *cobra.Command {
...
 			}
+			virtualNetworkCreateInput.SetVxlan(int32(vxlan))
+
 			if description != "" {
-				req.Description = description
+				virtualNetworkCreateInput.Description = &description
 			}
 
-			n, _, err := c.Service.Create(req)
+			n, _, err := c.Service.CreateVirtualNetwork(context.Background(), projectID).VirtualNetworkCreateInput(*virtualNetworkCreateInput).Include(nil).Exclude(nil).Execute()
```
ctreatma: There's no need to call `Include` or `Exclude` with a hard-coded `nil`; passing in `nil` is equivalent to not calling the method at all.

```suggestion
			n, _, err := c.Service.CreateVirtualNetwork(context.Background(), projectID).VirtualNetworkCreateInput(*virtualNetworkCreateInput).Execute()
```

internal/vlan/retrieve.go
```diff
@@ -41,19 +41,28 @@ func (c *Client) Retrieve() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			vnets, _, err := c.Service.List(projectID, c.Servicer.ListOptions(nil, nil))
+			request := c.Service.FindVirtualNetworks(cmd.Context(), projectID).Include(nil).Exclude(nil)
```
ctreatma: In this case, `Include` and `Exclude` need to be wired up to the command-line flags to match the old behavior

internal/vlan/retrieve.go
```diff
@@ -41,19 +41,28 @@ func (c *Client) Retrieve() *cobra.Command {
...
+			filters := c.Servicer.Filters()
+			if filters["facility"] != "" {
+				request = request.Facility(filters["facility"])
+			}
+
+			if filters["metro"] != "" {
+				request = request.Metro(filters["metro"])
+			}
+			VirtualNetworksList, _, err := request.Execute()
```
ctreatma: Variable names should start with a lower-case letter

test/e2e/vlan_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"strings"
+	"testing"
+
+	root "github.com/equinix/metal-cli/internal/cli"
+	outputPkg "github.com/equinix/metal-cli/internal/outputs"
+	"github.com/equinix/metal-cli/internal/vlan"
+	"github.com/spf13/cobra"
+)
+
+func TestCli_Users(t *testing.T) {
```
ctreatma: Test has the wrong name (mentions `Users`)

test/e2e/vlan_test.go
```diff
@@ -0,0 +1,66 @@
...
+		Outputer outputPkg.Outputer
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    *cobra.Command
+		cmdFunc func(*testing.T, *cobra.Command)
+	}{
+		{
+			name: "get",
```
ctreatma: Test fails because `project-id` flag is required but not provided

ctreatma (changes requested):

internal/vlan/retrieve.go
ctreatma: This was marked resolved but `Include` and `Exclude` are still not wired up to the command-line flags.

test/e2e/vlan_test.go
ctreatma: Note, for now you could set up the `project-id` flag by running `metal projects get` and taking the first project ID.

ctreatma (dismissed):

internal/vlan/retrieve.go
cprivitere: Looks like they're in there now.

cprivitere (approved):
Worked!

cprivitere:
Overriding test results cause it's due to being out of m3.smalls in Dallas.

PR #316, opened 2023-07-27
Updated metal-go client for sub-commands Users

codinja1188:


internal/users/retrieve.go
```diff
@@ -44,22 +45,22 @@ func (c *Client) Retrieve() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
 			var err error
-			var user *packngo.User
+			var user *metal.User
 			if userID == "" {
-				user, _, err = c.Service.Current()
+				user, _, err = c.Service.FindCurrentUser(context.Background()).Include(nil).Exclude(nil).Execute()
```
ctreatma: You can leave out the `Include` and `Exclude` calls here

internal/users/retrieve.go
```diff
@@ -44,22 +45,22 @@ func (c *Client) Retrieve() *cobra.Command {
...
 				if err != nil {
 					return fmt.Errorf("Could not get current User: %w", err)
 				}
 			} else {
-				user, _, err = c.Service.Get(userID, c.Servicer.ListOptions(nil, nil))
+				user, _, err = c.Service.FindUserById(context.Background(), userID).Include(nil).Exclude(nil).Execute()
```
ctreatma: This should be wired up to the `--include` and `--exclude` CLI flags to match the behavior of the old code

internal/users/retrieve.go
codinja1188: updated

internal/users/retrieve.go
codinja1188: updated

ctreatma: This is still passing in `nil` for include and exclude; it should be making a call to the servicer in order to pull in the values for the corresponding CLI flags.

codinja1188: updated

internal/users/add.go
```diff
@@ -46,20 +47,18 @@ func (c *Client) Add() *cobra.Command {
...
-				Invitee:     email,
-				Roles:       roles,
-				ProjectsIDs: projectsIDs,
-			}
-			invitation, _, err := c.InvitationService.Create(organizationID, createRequest, getOpts)
+			invitationInput := metal.NewInvitationInput(email)
+			invitationInput.SetRoles(roles)
+			invitationInput.SetProjectsIds(projectsIDs)
+
+			invitation, _, err := c.InvitationService.CreateOrganizationInvitation(context.Background(), organizationID).InvitationInput(*invitationInput).Execute()
```
ctreatma: It looks like this API supports `include` parameters: https://github.com/equinix-labs/metal-go/blob/main/metal/v1/api_organizations.go#L219

To match what the old code was doing with `getOpts`, this needs to be wired up to the `--include` command line flags via the servicer.

ctreatma (approved):

PR #319, opened 2023-07-30
docs: update contact info

jfreeland:
I was playing and was curious.  I believe the slack link has moved and IRC may no longer be valid.


ctreatma (approved):
Thanks for the PR!  You are correct on both counts.

Issue #265, opened 2023-05-12
metal init should work with either a Project or a User API token


displague:
`metal init` should work with either a Project or a User API token. If you use a project API token you won't be able to perform certain actions, but that may be desirable by the user.

One of the actions `metal init` performs is a User Preferences check to determine the default organization and project. That won't work with a project API token.

> https://github.com/equinix/metal-cli/blob/main/internal/init/init.go#L92
> `Error: Could not get current User: GET https://api.equinix.com/metal/v1/user: 403 Access denied for the current authentication token`

I'd propose an alternative lookup process during `metal init`:

- GET /projects with the token provided. limit the results to 2 (we only care if the count is == 1)
  - if there is an error, the token is invalid (ask again)
  - if there are 0, warn and skip project detection. (I don't believe metal-cli should offer to create an initial project for the user)
  - If there is 1 project:
    - just use that project as the default. we will also know the default organization by taking it from that project. (this is new code, iirc)
  - otherwise, we know this is a User API Key and we'll need to know which project should be default.

- fetch the User Preferences to discover the default project and org (this is existing code)
  - prompt the user with the project defaulted (this is existing code)

_Originally posted by @displague in https://github.com/equinix/metal-cli/issues/262#issuecomment-1544129367_

cprivitere:
We went with the above proposed algorithm, which has one notable shortcoming. If a user supplies a user API key, has multiple projects, and a default org with no projects in it, metal-cli will pick one of the orgs of those projects as the user's default organization rather than their actual default organization.

Fixing this entails changing the algorithm to instead assume all API keys are user API keys until the API gives a 403 response, at which point, we would treat it like a project API key.

This corner case is obscure enough that we're going to close this for now in favor of other higher priority work and can come back and re-open this later.

PR #325, opened 2023-08-08
feat: use both project and user API token and populate default org id and projectid

aayushrangwala:
#265

### DIFF
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -124,7 +124,7 @@ func (c *Client) NewCommand() *cobra.Command {
 			}
 
 			// Choose the first project in the preferred org
-			// Donot try to get projects again when
+			// Do not try to get projects again when
 			if defaultProjectId == "" && !isProjectToken {
 				defaultProjectId, err = getFirstProjectID(c.ProjectService, userOrg)
 				if err != nil {
@@ -158,6 +158,7 @@ func getDefaultProjectOK(service metal.ProjectsApiService) (*metal.Project, bool
 		return nil, isProjectToken, err
 	}
 
+	// Found more than one project, so must be a user token
 	if len(projects) > 1 {
 		return nil, isProjectToken, nil
 	}
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -145,9 +145,7 @@ func parseServiceToken(p metal.ProjectsApiService, u metal.UsersApiService) (str
 		// If only one project comes back, assume it's a project token
 		// and grab the ProjectID and Org Id from that single project
 		defaultProjectId = projects[0].GetId()
-		fmt.Println("Default Project ID: ", defaultProjectId)
 		defaultOrgId = path.Base(projects[0].Organization.AdditionalProperties["href"].(string))
-		fmt.Println("Default Org ID: ", defaultOrgId)
 	case numProj > 1:
 		// If more than one project comes back it must be a user token.
 		defaultOrgId, defaultProjectId, err = getDefaultIds(u)
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -103,11 +103,10 @@ func (c *Client) NewCommand() *cobra.Command {
 
 			if !isProjectToken {
 				// API token provided is user token
-				include := []string{} // []string | Nested attributes to include. Included objects will return their full attributes. Attribute names can be dotted (up to 3 levels) to included deeply nested objects. (optional)
 				exclude := []string{"devices", "members", "memberships", "invitations", "ssh_keys", "volumes", "backend_transfer_enabled", "updated_at", "customdata", "event_alert_configuration",
 					"timezone", "features", "avatar_url", "avatar_thumb_url", "two_factor_auth", "mailing_address", "max_projects", "verification_stage", "emails", "phone_number", "restricted",
 					"full_name", "email", "social_accounts", "opt_in_updated_at", "opt_in", "first_name", "last_name", "last_login_at"}
-				user, _, err := c.UserService.FindCurrentUser(context.Background()).Include(include).Exclude(exclude).Execute()
+				user, _, err := c.UserService.FindCurrentUser(context.Background()).Exclude(exclude).Execute()
 				if err != nil {
 					return err
 				}
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -25,6 +25,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


cprivitere (commented):
This seems to run much slower now than the current release. I assume that's due to it doing much more searching?

internal/init/init.go
```diff
@@ -91,34 +92,49 @@ func (c *Client) NewCommand() *cobra.Command {
 			c.UserService = *metalGoClient.UsersApi
 			c.ProjectService = *metalGoClient.ProjectsApi
 
-			user, _, err := c.UserService.FindCurrentUser(context.Background()).Execute()
+			var defaultOrgId, defaultProjectId string
+			project, isProjectToken, err := getDefaultProjectOK(c.ProjectService)
```
ctreatma: If the token is a user token, this has the potential to be extremely slow, because we're going to load every single project the user has access to before we look at the user's profile (which might already have a default project ID).  To reduce the impact on runtime for existing use cases, we could instead (pseudocode):

```
user, err := FindCurrentUser()

if err != nil {
  // I don't know what this error looks like; we might be able to make smart decisions based on error message
  projects, err := GetAllProjects()
  if err != nil {
    return err
  }
  if length(projects) > 1 {
    return errors.New("Failed to load user but got more than one project, I don't know what kind of token this is")
  } else {
    // Do project token setup
  }
}

// Successfully loaded current user, do user token setup
}
```

cprivitere:
So just tested this out, it's nearly 0 delay while waiting for it get your project id and org on the current main branch, and about 3-4 seconds on this new version. If we're ok with that, we are good to go.

internal/init/init.go
ctreatma: This PR needs to be rebased, as the codebase and its dependencies have changed quite a bit in the meantime (for example, the SDK now provides pagination helpers, so we don't need to (and shouldn't) use the `GetAllProjects` function), but in general the pseudocode above should allow us to add support for project tokens without drastically increasing the startup time when someone is using a user token.

internal/init/init.go
```diff
@@ -90,38 +91,53 @@ func (c *Client) NewCommand() *cobra.Command {
 			c.UserService = *metalGoClient.UsersApi
 			c.ProjectService = *metalGoClient.ProjectsApi
 
-			include := []string{} // []string | Nested attributes to include. Included objects will return their full attributes. Attribute names can be dotted (up to 3 levels) to included deeply nested objects. (optional)
-			exclude := []string{"devices", "members", "memberships", "invitations", "ssh_keys", "volumes", "backend_transfer_enabled", "updated_at", "customdata", "event_alert_configuration",
-				"timezone", "features", "avatar_url", "avatar_thumb_url", "two_factor_auth", "mailing_address", "max_projects", "verification_stage", "emails", "phone_number", "restricted",
-				"full_name", "email", "social_accounts", "opt_in_updated_at", "opt_in", "first_name", "last_name", "last_login_at"}
-			user, _, err := c.UserService.FindCurrentUser(context.Background()).Include(include).Exclude(exclude).Execute()
```
ctreatma, line 97: IMO we should stick with loading the current user first so that, if we _were_ given a user token, we avoid loading more projects than necessary or making multiple paginated calls to the projects endpoint.  What is the response and/or error that comes back if we try to run `c.UserService.FindCurrentUser(context.Background()).Include(include).Exclude(exclude).Execute()` with a project token?  Is there something in the response/error that we can use to decide that the token might be a project token?

displague: The description of #265 had an approach in mind

internal/init/init.go
```diff
@@ -90,38 +91,53 @@ func (c *Client) NewCommand() *cobra.Command {
...
-			project := user.GetDefaultProjectId()
-			fmt.Printf("Organization ID [%s]: ", organization)
+			if project != nil {
+				defaultProjectId = project.GetId()
+				defaultOrgId = project.Organization.GetId()
+			}
+
+			if !isProjectToken {
+				// API token provided is user token
+				include := []string{} // []string | Nested attributes to include. Included objects will return their full attributes. Attribute names can be dotted (up to 3 levels) to included deeply nested objects. (optional)
```
ctreatma: We can take out the `.Include()` entirely if we're not sending any include params.

internal/init/init.go
ctreatma: This PR is already largely following the steps in the issue description.  I still prefer adding project API support in a way that does not add latency to user token initialization, but adding the 2-project limit described in that issue to this PR would at least reduce the added latency.

ctreatma: @Chris pointed out that a user might only have 1 project, in which case `metal init` would treat their token as a project token instead of a user token; the token type doesn't have any impact outside of the init command, though, so which endpoint we try to load first is largely a philosophical issue.  I leave the decision up to the contributor as to whether we should remove auto-pagination from `getProjectsOK` or load the current user endpoint first and then try projects if current user 403s.

internal/init/init.go
cprivitere: Removed.

internal/init/init.go
```diff
@@ -135,22 +124,60 @@ func (c *Client) NewCommand() *cobra.Command {
...
+		fmt.Println("WARN: No available projects found with the provided API Token")
+		defaultOrgId, defaultProjectId, err = getDefaultIds(u)
+		if err != nil {
+			return "", "", err
 		}
+	case numProj == 1:
+		// If only one project comes back, assume it's a project token
+		// and grab the ProjectID and Org Id from that single project
+		defaultProjectId = projects[0].GetId()
+		fmt.Println("Default Project ID: ", defaultProjectId)
```
displague: These `fmt.Println`s are redundant to the prompts which include the defaults..

```
$ metal init
Equinix Metal API Tokens can be obtained through the portal at https://console.equinix.com/profile/api-keys
See https://metal.equinix.com/developers/docs/accounts/users/ for more details.

User Token (hidden): 
Default Project ID:  248...
Default Org ID:  ecd...
Organization ID [ecd...]: 
Project ID [248...]: 

Writing /home/codespace/.config/equinix/metal.yaml
```

displague (dismissed):
I left a nitpick about the output which seems to be more of a debugging aid rather than necessary or helpful to end-users. https://github.com/equinix/metal-cli/pull/325/files#r1471550998 

If this text were accompanied by a full list of organization ids that the key could access, it may be more useful. I would want to defer that work to accompany investigation into TUI or more interactive prompting tools (with name based autocompletes or console drop downs, for example).

Otherwise, this has been tested for a User API Key and Project API key. In my User testing I had a null project id default for some tests and a defined id - worked fine under both conditions.

internal/init/init.go
cprivitere: That was debug output. I've removed it.

displague (approved):

Issue #286, opened 2023-06-16
`metal hardware-reservations` does not include metros


displague:
### What happened?

https://eqix-metal-community.slack.com/archives/C4FEV702E/p1686913291297669

> right now to check if we have a reserved node available at a certain metro we run `metal-cli hardware-reservations get -o json`  and then parse the output with `jq` to select provisionable nodes at a certain metro like so:  `select(.facility.metro.code==\"$METRO\" and .provisionable==true)` . The potential issue here is that we get the metro code from the facility schema, is there a better way to do this that doesn't involve going through the facility schema in json? I am running metal-cli version 0.14.1.

### What should happen?

The Metal CLI should expose the Metro field alongside the facility field.

aayushrangwala:
@displague SInce this issue is fixed, can you please close it

PR #326, opened 2023-08-08
Provide metro as well with facility code for hardware reservation

aayushrangwala:
#286

### DIFF
--- a/internal/hardware/retrieve.go
+++ b/internal/hardware/retrieve.go
@@ -44,9 +44,9 @@ func (c *Client) Retrieve() *cobra.Command {
 			projectID, _ := cmd.Flags().GetString("project-id")
 			hardwareReservationID, _ := cmd.Flags().GetString("id")
 
-			header := []string{"ID", "Facility", "Plan", "Created"}
+			header := []string{"ID", "Facility", "Metro", "Plan", "Created"}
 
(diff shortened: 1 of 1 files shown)


internal/hardware/retrieve.go
```diff
@@ -44,9 +44,9 @@ func (c *Client) Retrieve() *cobra.Command {
 			projectID, _ := cmd.Flags().GetString("project-id")
 			hardwareReservationID, _ := cmd.Flags().GetString("id")
 
-			header := []string{"ID", "Facility", "Plan", "Created"}
+			header := []string{"ID", "Facility", "Metro", "Plan", "Created"}
 
-			inc := []string{}
+			inc := []string{"facility"}
```
codinja1188: Is it included in standard cases too? If, Yes, need to remove it. 

```
                        case outputs.FormatJSON, outputs.FormatYAML:
-                               inc = append(inc, "project", "facility", "device")
+                               inc = append(inc, "project", "device")
                        }
```

aayushrangwala: Yes, updated the switch case to consider `facility.metro` only as we only need that much.

cprivitere (approved):
Tried it out with Ayush's reservation and it looks good!

PR #282, opened 2023-06-14
Updated metal-go client for sub-command devices

codinja1188:
Breakout from https://github.com/equinix/metal-cli/pull/270

What this PR does / why we need it:

This PR replaces packngo with metal-go for all interactions with the Equinix Metal API specifically for Device sub commands

Test Results:
================
# Retrieve
## Cmd: metal devices get

* Packngo:
```
vasu@vasu-cloud:~/Equinix$ metal devices get
+--------------------------------------+-------------------------------------+------------------+--------+----------------------+
|                  ID                  |              HOSTNAME               |        OS        | STATE  |       CREATED        |
+--------------------------------------+-------------------------------------+------------------+--------+----------------------+
| d231a5d0-3da9-4e4b-bbd6-a367c4c2d34b | test-staging-2                      | Ubuntu 20.04 LTS | active | 2023-06-01T08:42:42Z |
| bcab4df0-945e-4cb5-a8c7-f9dfbaabb3d7 | eksa-node-003                       | Custom iPXE      | active | 2023-04-25T12:33:17Z |
| 09be98b7-0db6-4fdb-90be-901ab847158d | test-packngo                        | Ubuntu 20.04 LTS | active | 2023-06-26T13:09:24Z |
| b396c4fd-c412-4a25-b6b4-df6f2f548629 | test-staging-12                     | Ubuntu 22.04 LTS | active | 2023-06-01T08:43:12Z |
| bad6f156-4777-49a5-87e7-d038e7365c9e | capi-quickstart-control-plane-mkz7l | Ubuntu 22.04 LTS | active | 2023-05-03T09:42:52Z |
| d7126b67-a96f-48b7-9cf1-e1914d84235b | capi-quickstart-worker-a-spr64      | Ubuntu 22.04 LTS | active | 2023-05-03T09:48:01Z |
| 4c397b2d-eccf-4948-9fcf-2b325303048c | capi-quickstart-control-plane-98xjq | Ubuntu 22.04 LTS | active | 2023-05-03T09:48:05Z |
| bdbe09fa-d140-4cc9-b988-f51610e8c2f3 | capi-quickstart-worker-a-q89f8      | Ubuntu 22.04 LTS | active | 2023-05-03T09:47:59Z |
| 67992861-066b-4d76-a6a4-a21b3e855722 | capi-quickstart-worker-a-qqnhb      | Ubuntu 22.04 LTS | active | 2023-05-03T09:48:00Z |
[...]

PR #330, opened 2023-08-11
chore(ci): update e2e test workflow to allow nested directories

ctreatma:
The end-to-end tests for #282 are failing because the tests were moved to nested directories.  This updates the e2e test workflow to not care about how the tests are laid out within the test/e2e directory so that we can move things around freely.


codinja1188 (approved):
/lgtm

PR #334, opened 2023-08-16
chore: Updated metal-go client for sub-command twofa

codinja1188:

This PR replaces packngo with metal-go for all interactions with the Equinix Metal API specifically for twofa subcommands


internal/twofa/receive.go
```diff
@@ -1,75 +0,0 @@
...
-				_, err := c.Service.ReceiveSms()
-				if err != nil {
-					return fmt.Errorf("Could not issue token via SMS: %w", err)
-				}
-
-				fmt.Println("SMS token sent to your phone")
-				return nil
-			}
-
-			otpURI, _, err := c.Service.SeedApp()
```
ctreatma, line 59: Is this subcommand no longer needed?  Or is it removed because the endpoints/operations used here are not in the spec?

If we still need to support this subcommand, a short-term workaround might be to find a way to convert the other 2fa subcommands to metal-go while leaving this one on packngo.

internal/twofa/enable2fa.go
```diff
@@ -48,12 +49,13 @@ func (c *Client) Enable() *cobra.Command {
 
 			cmd.SilenceUsage = true
 			if sms {
-				_, err := c.Service.EnableSms(token)
+				_, err := c.Service.EnableTfaSms(context.Background()).Execute()
```
ctreatma, line 52: Do we need to be able to pass in the token?  Is this a spec issue that needs to be fixed?

aayushrangwala:
@vasubabu Please decline this PR in the favour of https://github.com/equinix/metal-cli/pull/353

ctreatma:
Closing in favor of #353

PR #338, opened 2023-08-30
chore: add workflow to enforce Conventional Commit tags

ctreatma:
Currently our release process is fairly manual:  someone with write access to this repo has to check out the latest from `main`, look at what has changed since the last release, decide on an appropriate new version number, create a git tag in the correct format (is it `x.y.z` or `vx.y.z`?), and push that tag back up to GitHub to trigger the release process.

By ensuring that our PR titles have Conventional Commit tags, we make it possible to automatically determine the next release version number, which means we can move to a push-button release process: click the release button, and the workflow does the rest.

This workflow has been in place for some time on `metal-go` and was recently adopted in `metal-java`; adopting it more broadly will make it easier for us to safely share the responsibility for releasing our various tools (and could even lead to automated / scheduled release workflows instead of push-button workflows).

### DIFF
--- a/.github/workflows/validate_pr.yml
+++ b/.github/workflows/validate_pr.yml
@@ -1,0 +1,45 @@
+name: "Validate PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: write
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        id: lint_pr_title
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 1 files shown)


ctreatma:
This workflow checks the PR title, not the commit messages, so adding the Conventional Commit tag can be done by editing the PR title without making any changes to the commits in the PRs.

When a PR is opened without Conventional Commit tags, the workflow adds a comment to the PR that explains what the issue is and how to fix it (screenshot below is taken from [equinix-labs/metal-go#134](https://github.com/equinix-labs/metal-go/pull/134#issuecomment-1699487774):
Screenshot 2023-08-31 at 9 45 02 AM

When the PR title is updated to fix the issue, the comment is deleted.  By enforcing tags on the PR titles, we make it possible for internal people to add the necessary tags if, for example, we get a good PR from an external contributor and that contributor is not motivated to add the necessary tag or picks the wrong tag.

displague (approved):

Issue #257, opened 2023-03-08
Resetting always_pxe flag does not work


lheckemann:
### What happened?
I ran `metal device update -i $instance_uuid --always-pxe=false`, and the always_pxe flag was still set to true afterwards.

### How can we reproduce it?

Run `metal device update -i $instance_uuid --always-pxe=false` and check the instance config afterwards.

CLI version (`metal --version`): metal version 0.13.0

Workaround: `curl -H @<(echo "X-Auth-Token: $METAL_AUTH_TOKEN") https://api.equinix.com/metal/v1/devices/$instance_uuid -H 'Content-Type: application/json' -X PUT -d @<(jq -n '{always_pxe: false}')`

grahamc:
ref: https://github.com/equinix/metal-cli/blob/55108842ba065f14050da24e103038cb87c54ad4/internal/devices/update.go#L76-L78

PR #340, opened 2023-08-31
test: Improved device unit test cases and fixed issue

codinja1188:
## Description:

We improved the device unit cases

## Issue fixed?
https://github.com/equinix/metal-cli/issues/257


test/helper/helper.go
```diff
@@ -74,6 +74,29 @@ func IsDeviceStateActive(deviceId string) (bool, error) {
 	return false, err
 }
 
+func IsDeviceStateInActive(deviceId string) (bool, error) {
```
ctreatma: Rather than create a separate function, the existing helper could be updated to accept a `state` as an argument; that way we can avoid creating a new function if we find a third or fourth state we need to wait for.

test/helper/helper.go
```diff
@@ -74,6 +74,29 @@ func IsDeviceStateActive(deviceId string) (bool, error) {
...
+			return false, err
+		}
+		if resp.GetState() == "inactive" {
+			return true, nil
+		}
+
+		// Sleep for the specified interval
+		time.Sleep(retryInterval)
+	}
+	return false, err
```
ctreatma: Note that this is going to return a `nil` error; instead we should return an error that describes what happened.  #339 shows how we could do that

test/helper/helper.go
codinja1188: updated

test/helper/helper.go
codinja1188: updated

test/helper/helper.go
```diff
@@ -51,7 +51,9 @@ func CreateTestDevice(projectId, name string) (string, error) {
 	return deviceResp.GetId(), nil
 }
 
-func IsDeviceStateActive(deviceId string) (bool, error) {
+func IsDeviceStateActive(deviceId string, state string) (bool, error) {
+	var err error
```
cprivitere:
```suggestion
```

test/helper/helper.go
```diff
@@ -51,7 +51,9 @@ func CreateTestDevice(projectId, name string) (string, error) {
...
+	var resp *openapiclient.Device
```
cprivitere:
```suggestion
```

cprivitere:
Was looking at this, seems like the state shouldn't be a string but instead some type defined in the generated API files.

internal/devices/create.go
```diff
@@ -125,8 +125,13 @@ func (c *Client) Create() *cobra.Command {
 				if billingCycle != "" {
 					facilityDeviceRequest.DeviceCreateInFacilityInput.SetBillingCycle(*validBillingCycle)
 				}
-				if alwaysPXE {
-					facilityDeviceRequest.DeviceCreateInFacilityInput.SetAlwaysPxe(alwaysPXE)
+				if alwaysPXE != "" {
```
displague, line 128: These changes were addressed in https://github.com/equinix/metal-cli/pull/282#pullrequestreview-1550002522

https://github.com/equinix/metal-cli/pull/282/files#diff-6153c670828ec05635bb32b940eab661ce8225f32ca7d5152b9c4d1b8576e1b9R116-R119

displague (commented):
Is anything left in this PR that hasn't been addressed in another PR? Anything that this PR aimed to do that it hasn't done?

Should we close this?

ctreatma:
Closing this since everything in here appears to have been addressed in other PRs.

PR #350, opened 2023-09-20
chore: upgrade metal-go to 0.22.2 and fix enum type errors

ctreatma:
The latest release of metal-go introduced true enum fields.  Previously, metal-go generated enum fields as plain strings due to a bug in openapi-generator.  Now that enum fields are generated as Go enums, we need to convert user-provided string values for those fields to the correct enum type for input and we need to convert the enum values to plain strings for output.

This updates metal-go to the latest release and addresses type errors for metal-go client code as it exists in the `main` branch; PRs that introduce metal-go elsewhere in the CLI may need to be updated as well before they can be merged.

### DIFF
--- a/internal/devices/create.go
+++ b/internal/devices/create.go
@@ -97,6 +97,17 @@ func (c *Client) Create() *cobra.Command {
 			var facilityArgs []string
 
 			var request metal.ApiCreateDeviceRequest
+
+			var validBillingCycle *metal.DeviceCreateInputBillingCycle
+			var err error
+
+			if billingCycle != "" {
+				validBillingCycle, err = metal.NewDeviceCreateInputBillingCycleFromValue(billingCycle)
+				if err != nil {
+					return err
+				}
+			}
+
 			if facility != "" {
 				facilityArgs = append(facilityArgs, facility)
 
@@ -112,7 +123,7 @@ func (c *Client) Create() *cobra.Command {
 				}
 
 				if billingCycle != "" {
-					facilityDeviceRequest.DeviceCreateInFacilityInput.SetBillingCycle(billingCycle)
+					facilityDeviceRequest.DeviceCreateInFacilityInput.SetBillingCycle(*validBillingCycle)
 				}
 				if alwaysPXE {
 					facilityDeviceRequest.DeviceCreateInFacilityInput.SetAlwaysPxe(alwaysPXE)
@@ -151,7 +162,7 @@ func (c *Client) Create() *cobra.Command {
 					},
 				}
 				if billingCycle != "" {
-					metroDeviceRequest.DeviceCreateInMetroInput.SetBillingCycle(billingCycle)
+					metroDeviceRequest.DeviceCreateInMetroInput.SetBillingCycle(*validBillingCycle)
 				}
 
 				if alwaysPXE {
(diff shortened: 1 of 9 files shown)


internal/organizations/retrieve.go
```diff
@@ -44,12 +44,11 @@ func (c *Client) Retrieve() *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			include := []string{"Inner_example"}
-			exclude := []string{"Inner_example"}
-			withoutProjects := "withoutProjects_example"
```
ctreatma, line 49: The `withoutProjects` option did not exist prior to `metal-go` conversion, and is not wired up to any user-facing flags, so I've opted to remove it rather than fix the type error for now.

internal/pagination/pager.go
```diff
@@ -108,13 +108,13 @@ func GetAllEvents(s metal.ApiFindEventsRequest) ([]metal.Event, error) {
 	}
 }
 
-func GetAllOrganizations(s metal.OrganizationsApiService, include, exclude []string, withOutProjects string) ([]metal.Organization, error) {
+func GetAllOrganizations(s metal.OrganizationsApiService, include, exclude []string) ([]metal.Organization, error) {
```
ctreatma, line 111: Removing the `withoutProjects` setting from the `organization get` subcommand means we can also remove it here.

internal/plans/retrieve.go
```diff
@@ -43,7 +44,11 @@ func (c *Client) Retrieve() *cobra.Command {
 			filters := c.Servicer.Filters()
 
 			if filters["type"] != "" {
-				request = request.Type_(filters["type"])
+				validType, err := metal.NewFindPlansTypeParameterFromValue(filters["type"])
+				if err != nil {
+					return err
```
ctreatma, line 49: I opted to just return the error here rather than wrapping it, since the error that's thrown is already helpful: https://github.com/equinix-labs/metal-go/blob/main/metal/v1/model_plan_type.go#L60

internal/users/add.go
```diff
@@ -74,3 +82,15 @@ func (c *Client) Add() *cobra.Command {
 
 	return addUserCmd
 }
+
+func validateRoles(roles []string) ([]metal.InvitationRolesInner, error) {
+	validRoles := make([]metal.InvitationRolesInner, len(roles))
+	for _, role := range roles {
+		validRole, err := metal.NewInvitationRolesInnerFromValue(role)
+		if err != nil {
+			return nil, err
```
ctreatma, line 91: Again opted to return the original error without wrapping it, since the message looks good to me: https://github.com/equinix-labs/metal-go/blob/main/metal/v1/model_invitation_roles_inner.go#L62

internal/devices/create.go
```diff
@@ -97,6 +97,17 @@ func (c *Client) Create() *cobra.Command {
...
 
 			var request metal.ApiCreateDeviceRequest
+
+			var validBillingCycle *metal.DeviceCreateInputBillingCycle
+			var err error
+
+			if billingCycle != "" {
+				validBillingCycle, err = metal.NewDeviceCreateInputBillingCycleFromValue(billingCycle)
+				if err != nil {
+					return err
```
ctreatma, line 107: This is returning the original error because that message is nicely descriptive: https://github.com/equinix-labs/metal-go/blob/main/metal/v1/model_device_create_input_billing_cycle.go#L52

The error message does make mention of the type `DeviceCreateInputBillingCycle`; it would be preferable to reference either the CLI flag name or the attribute name there, but IMO tweaking that is not worth the effort at the moment.

ctreatma:
The API spec for the plan type enum has `NVME`, but the API response has `NVMe`, so the plan subcommand is broken by the introduction of true enums: https://github.com/equinix/metal-cli/actions/runs/6252218404/job/16974991464#step:5:116

ctreatma:
There were other data validation issues that I missed somehow when running these tests locally; another `metal-go` PR is up to remove a couple inaccurate enums from the plan spec: https://github.com/equinix-labs/metal-go/pull/156

codinja1188 (approved):
/LGTM

ctreatma:
Updated to run with `metal-go@v0.22.2` and [the end-to-end tests are passing again](https://github.com/equinix/metal-cli/actions/runs/6265599925/job/17014783361?pr=350)

cprivitere (approved):

PR #362, opened 2023-10-07
feat: BETA loadbalancer service support

displague:
Part of https://github.com/equinix/metal-cli/issues/411

(Used CPEM's LBaaS Go project as the base. Lots of cleanup ahead. )

This PR introduces the minimal subcommand support to start working with LoadBalancers. This PR does not include ports and pools, which can be added later.

TODO:
* [x] Get LB
* [x] List LB
* [x] Detele LB
* [x] Create Project LB
* [x] Send debug transport to token exchange when debugging?
* [x] Ensure command docs are correct
* [ ] Ensure requirement checks (required fields) are correct
* [x] Ensure example commands are correct
* [x] Ensure command is named "loadbalancer-beta"
* [x] generate docs


docs/metal.md
```diff
@@ -38,6 +38,7 @@ Command line interface for Equinix Metal
 * [metal init](metal_init.md)	 - Create a configuration file.
 * [metal interconnections](metal_interconnections.md)	 - interconnections operations: create, get, update, delete
 * [metal ip](metal_ip.md)	 - IP address, reservations, and assignment operations: assign, unassign, remove, available, request and get.
+* [metal loadbalancer-beta](metal_loadbalancer-beta.md)	 - LoadBalancer BETA operations: create, get, update, delete, and bgpconfig.
```
tylerauerbeck: The `metal_loadbalancer-beta.md` doesn't appear to exist. And I think we can drop the `bgpconfig` bit from this line

internal/loadbalancers/infrastructure/manager.go
```diff
@@ -0,0 +1,166 @@
...
+	"fmt"
+	"reflect"
+
+	lbaas "github.com/equinix/metal-cli/internal/loadbalancers/api/v1"
+)
+
+const ProviderID = "loadpvd-gOB_-byp5ebFo7A3LHv2B"
+
+var LBMetros = map[string]string{
+	"da": "lctnloc--uxs0GLeAELHKV8GxO_AI",
```
tylerauerbeck: Let's make sure to get DC added in here `lctnloc-1lJjVT6Zp_Fs4UpW5LWQu` while we sort out how to not manage this manually.

internal/loadbalancers/create.go
```diff
@@ -0,0 +1,93 @@
...
+			header := []string{"ID", "Message", "Version", "Status"}
+			return c.Out.Output(lb, header, &data)
+		},
+	}
+
+	createLoadBalancerCmd.Flags().StringVarP(&name, "name", "n", "", "Name of the loadbalancer")
+	createLoadBalancerCmd.Flags().StringVarP(&projectID, "project-id", "p", "", "The project's UUID. This flag is required, unless specified in the config created by metal init or set as METAL_PROJECT_ID environment variable.")
+	createLoadBalancerCmd.Flags().StringVarP(&locationId, "location", "l", "", "The location's ID.")
+	createLoadBalancerCmd.Flags().StringVarP(&providerId, "provider", "r", ProviderID, "The provider ID.")
+	createLoadBalancerCmd.Flags().StringSliceVar(&portIds, "port", []string{}, "The port's UUID. This flag is required, unless specified in the config created by metal init or set as METAL_PORT_ID environment variable.")
```
tylerauerbeck: Is `METAL_PORT_ID` an existing construct? I guess more curious why this one has an environment variable and the others wouldn't if that's not the case.

internal/loadbalancers/delete.go
```diff
@@ -0,0 +1,85 @@
...
+			return err
+		}
+		fmt.Println("LoadBalancer", loadbalancerID, "successfully deleted.")
+		return nil
+	}
+	// deleteLoadBalancerCmd represents the deleteLoadBalancer command
+	deleteLoadBalancerCmd := &cobra.Command{
+		Use:   `delete --id <loadbalancer_UUID> [--force]`,
+		Short: "Deletes a loadbalancer.",
+		Long:  "Deletes the specified loadbalancer with a confirmation prompt. To skip the confirmation use --force. You can't delete a loadbalancer that has active resources. You have to deprovision all pools and ports from a loadbalancer in order to delete it.",
```
tylerauerbeck: So when a loadbalancer is deleted, it will delete the associated ports. But this is correct in that pools and origins need cleaned up separately

tylerauerbeck (changes requested):
Just a few comments, but overall this looks like a great start!

internal/loadbalancers/api/v1/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
ctreatma, line 1: The internal client in CPEM was made to hit a deadline.  We don't have that time pressure here, so this should use a generated client that lives in [equinix-sdk-go](https://github.com/equinix/equinix-sdk-go)

internal/loadbalancers/infrastructure/token_exchanger.go
```diff
@@ -0,0 +1,67 @@
...
+		client = http.DefaultClient
+	}
+	return &TokenExchanger{
+		metalAPIKey: metalAPIKey,
+		client:      client,
+	}
+}
+
+func (m TokenExchanger) Token() (*oauth2.Token, error) {
+	tokenExchangeURL := "https://iam.metalctrl.io/api-keys/exchange"
```
ctreatma, line 31: Don't know that it would add much here, but if we had a spec for iam.metalcrtl.io we could use a generated client.

internal/loadbalancers/infrastructure/manager.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+type Manager struct {
+	client         *lbaas.APIClient
+	metro          string
+	projectID      string
+	tokenExchanger *TokenExchanger
+}
+
+func NewManager(metalAPIKey, projectID, metro string) *Manager {
```
ctreatma: In CPEM, the `Manager` type was intended to abstract away the concerns in that particular application, where we need to reconcile a desired load balancer state in k8s with what exists in the world.  In the CLI, there's no reconciliation, so we should use the LB client directly the same way we do for Metal.

docs/metal.md
displague: 🤦 `git add` ... 
`bgpconfig` is a copy/paste from using the Project subcommand as the base. I tried to catch these but looks like I missed some.

internal/loadbalancers/create.go
displague: Technically, every parameter can be supplied as an environment variable or a ~/.config/equinix/metal.yaml setting.

We probably don't need to be explicit here about METAL_PORT_ID (I would guess I copied this from the Project subcommand and made minor adjustments to make it fit).

internal/loadbalancers/infrastructure/token_exchanger.go
displague: cc @tylerauerbeck

internal/loadbalancers/infrastructure/manager.go
displague: I copied the directory wholesale. Can remove it from this PR.

internal/loadbalancers/create.go
displague: Port isn't required in any case and this comment should have been applied to location. Fixing.

internal/loadbalancers/delete.go
displague: Got it. Pools and origins are independent. ports do not need to be deleted in advance.

internal/loadbalancers/infrastructure/manager.go
ctreatma: Assuming that the LB client gets moved over to `equinix-sdk-go`, the `ProviderID` constants and `LBMetros` var should be moved to the LB client (but as mentioned in a different comment, the Manager type should not be part of the shared client b/c CLI users will likely want more granular operations than the ones we implemented for CPEM).

ctreatma:
One thing to consider here: if LBaaS has its own generated client that sits next to the metal client in `equinix-sdk-go`, with its own spec and its own API endpoint, that may indicate that LBaaS CLI commands should go alongside Metal CLI commands in a broader-scoped Equinix CLI (`eqx`? `eqix`?  `the-equinix-cli`?) rather than directly in here.

internal/loadbalancers/api/v1/.gitignore
cprivitere: Seems like we decided to go with the internal clients for now.

internal/loadbalancers/infrastructure/manager.go
cprivitere: Seems like we decided to go with the internal clients for now.
Also, dc is in here now.

cprivitere:
Merged in latest updates and tested create/get/delete commands to verify they work.

cprivitere (commented):
> One thing to consider here: if LBaaS has its own generated client that sits next to the metal client in `equinix-sdk-go`, with its own spec and its own API endpoint, that may indicate that LBaaS CLI commands should go alongside Metal CLI commands in a broader-scoped Equinix CLI (`eqx`? `eqix`?  `the-equinix-cli`?) rather than directly in here. 

I'm all for dropping metal-cli and making eq-cli or whatever.

PR #365, opened 2023-10-16
fix: improved init command excluded unnecessary flags

codinja1188:
# Description 

Excluded unnecessary flags in "metal init"


cprivitere (approved):

cprivitere:
This is much faster, nice!

PR #367, opened 2023-10-17
fix: For Resetting always-pxe flag

codinja1188:
# Description :

This PR  fixes the issue https://github.com/equinix/metal-cli/issues/257

PR #370, opened 2023-10-24
feat: add --userdata-file to device/update

dlotterman:
Some preferential doc sting changes to highlight the ease of the recent [changes to updating always-pxe](https://github.com/equinix/metal-cli/commit/a64058f61e60431df65e920a6d705aac9511cd1b)

Also C+V, C+V's the `--userdata-file` code from device/create to device/update, making it easier to update complicated userdata formats for devices in place.

```
$ bin/metal device update -i SOMEDEVICEUUID --userdata-file /tmp/my-cloud-init.mime
+--------------------------------------+----------+-------------+--------+
|                  ID                  | HOSTNAME |     OS      | STATE  |
+--------------------------------------+----------+-------------+--------+
| 9ffac3bb-4e04-466b-a786-b703fe52fa7f | h-01     | Custom iPXE | active |
+--------------------------------------+----------+-------------+--------+
```

Passed `make generate-docs` 🤷 ?


displague:
@dlotterman I'm wondering if this PR was motivated by a bug that was recently addressed: https://github.com/equinix/metal-cli/pull/367#issuecomment-1768988846

A bare `--always-pxe` should have the behavior of enabling always-pxe, just as a bare `--locked` should lock the node.

When we were revisiting the Always PXE behavior, we did not look into if the Lock behavior had also been affected. (@cc @vasubabu)

internal/devices/update.go
```diff
@@ -62,6 +64,18 @@ func (c *Client) Update() *cobra.Command {
 				deviceUpdate.Description = &description
 			}
 
+			if userdata != "" && userdataFile != "" {
+				return fmt.Errorf("either userdata or userdata-file should be set")
```
displague, line 68: This change would require userdata changes whenever updating a device. 

"Want to change the hostname? Include userdata, please."

displague: PR otherwise looks good - just remove this condition block.

ctreatma: This won't force userdata changes; it's validating that the user only specified one of `--userdata <some_string>` or `--userdata-file <some_file>`.

dlotterman: This was my understanding as well ^^, just checking that both are not empty, or put another way if both are populated, aka one or less is specified?

So then this is using this branchs binary updating a device without updating userdatas:

```
$ ~/code/github/metal-cli/bin/metal device update -i 3b91c052-f218-4e22-a95a-4c9f057034c4 -t "metalcli,charlierulesok"
+--------------------------------------+----------+-------------+--------+
|                  ID                  | HOSTNAME |     OS      | STATE  |
+--------------------------------------+----------+-------------+--------+
| 3b91c052-f218-4e22-a95a-4c9f057034c4 | h-11     | Custom iPXE | active |
+--------------------------------------+----------+-------------+--------+
`

displague (approved):

Issue #387, opened 2023-11-09
TestPorts_VLANs E2E test failed


displague:
One E2E test failed:
```
=== RUN   TestPorts_VLANs
Error when calling `DevicesApi.FindDeviceById`: 403 Forbidden
    helper.go:302: 403 Forbidden
    helper.go:305: Timeout while waiting for device: 467ae367-bb48-469d-b910-8e71533cf42f to be active
Error when calling `DevicesApi.FindDeviceById`: 403 Forbidden
    helper.go:310: 403 Forbidden
--- FAIL: TestPorts_VLANs (65.85s)
panic: runtime error: index out of range [2] with length 0 [recovered]
	panic: runtime error: index out of range [2] with length 0

goroutine 36 [running]:
testing.tRunner.func1.2({0xca9f80, 0xc0001cb338})
	/opt/hostedtoolcache/go/1.19.13/x64/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.13/x64/src/testing/testing.go:1399 +0x39f
panic({0xca9f80, 0xc0001cb338})
	/opt/hostedtoolcache/go/1.19.13/x64/src/runtime/panic.go:884 +0x212
github.com/equinix/metal-cli/test/e2e/ports.TestPorts_VLANs(0xc0004f4000)
	/home/runner/work/metal-cli/metal-cli/test/e2e/ports/vlans_test.go:28 +0x4dc
testing.tRunner(0xc0004f4000, 0xd6b4f0)
	/opt/hostedtoolcache/go/1.19.13/x64/src/testing/testing.go:1446 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.13/x64/src/testing/testing.go:1493 +0x35f
FAIL	github.com/equinix/metal-cli/test/e2e/ports	596.680s
```

_Originally posted by @displague in https://github.com/equinix/metal-cli/issues/355#issuecomment-1804428216_

aayushrangwala:
Also in the latest test run https://github.com/equinix/metal-cli/actions/runs/6932109723/job/18860656801?pr=376

This specific test is passing
```
=== RUN   TestPorts_VLANs
=== RUN   TestPorts_VLANs/vlan_assignment_port
--- PASS: TestPorts_VLANs (223.85s)
    --- PASS: TestPorts_VLANs/vlan_assignment_port (15.80s)
PASS
```

[...]

PR #390, opened 2023-11-15
fix: output for convert layer2 to layer 3 command

### LINKED ISSUE #386: converting from layer2 to layer3 bond with IPV4 and IPV6 shows some junk output
why port convert from layer2 to layer3 bond with IPV4 and IPV6 shows some Junk data

```
Vasubabus-MacBook-Pro:bin vasubabu$ ./metal port convert -i d414a2ca-2e91-4e2d-a598-42180c354e9d -2=false -b -4 -6
2023/11/07 23:29:57 Converting port d414a2ca-2e91-4e2d-a598-42180c354e9d to layer-3 with addresses [{0x140002c2540 0x140002c2544 map[]} {0x140002c2540 0x140002c2544 map[]} {0x140002c2540 0x140002c2544 map[]}]
```

_Originally posted by @vasubabu in https://github.com/equinix/metal-cli/pull/355#pullrequestreview-1718403693_

aayushrangwala:
metal-cli Convert subcommand from layer 2 to layer 3 logs pointers for the addresses

Fixes: https://github.com/equinix/metal-cli/issues/386 and https://github.com/equinix/metal-cli/issues/387

### DIFF
--- a/internal/ports/convert.go
+++ b/internal/ports/convert.go
@@ -95,7 +95,7 @@ func (c *Client) Convert() *cobra.Command {
 			}
 
 			convToL3 := func(portID string) (*metal.Port, *http.Response, error) {
-				log.Printf("Converting port %s to layer-3 with addresses %v", portID, addrs)
+				log.Printf("Converting port %s to layer-3", portID)
 				return c.PortService.
 					ConvertLayer3(context.Background(), portID).
 					PortConvertLayer3Input(metal.PortConvertLayer3Input{


aayushrangwala:
@displague @ctreatma Please trigger E2E tests

internal/ports/convert.go
```diff
@@ -87,15 +90,20 @@ func (c *Client) Convert() *cobra.Command {
...
 				public = true
 				addrs = append(addrs, metal.PortConvertLayer3InputRequestIpsInner{AddressFamily: &addressFamily, Public: &public})
+
+				addressType = "public"
+				addressFamilyStrs = append(addressFamilyStrs, strconv.Itoa(int(metal.IPADDRESSADDRESSFAMILY__6)))
 			}
 
 			convToL3 := func(portID string) (*metal.Port, *http.Response, error) {
-				log.Printf("Converting port %s to layer-3 with addresses %v", portID, addrs)
+				log.Printf("Converting port %s to layer-3 with [%s] addresses IPv[%v]", portID, addressType, addressFamilyStrs)
```
displague: addressType will be `public` when both public and private (required, will always be true when enabling layer3).

Was the intent to say (pseudo-code):
`"Converting port %s to layer-3 with private IPv4" (conditionally " and public " (conditionally join(IPv4AndOrIPv6, ", ")) " addresses")`

displague: I don't think we offer this level of detail for anything else being created. We don't say, for example, "Creating a device in metro sv with plan c3.small.x86 and OS alpine"

I think we could omit this log message altogether if we really don't do it in the usual create or update case.
Perhaps a reason to keep it would be if the arguments to this command are so complicated that you never really know what to expect on the other end when you ask for something, but I think we have a handle on that now that we are addressing the bonding and converting actions not performing the actions that they logically should have.

aayushrangwala: Got it. Fixed

displague (approved):

Issue #331, opened 2023-08-14
Remove promptui library and calls


cprivitere:
We use https://github.com/manifoldco/promptui in this project which seems to be [abandoned](https://github.com/manifoldco/promptui/issues/222). We should probably replace this code with something else.

cprivitere:
https://github.com/search?q=repo%3Aequinix%2Fmetal-cli+%22github.com%2Fmanifoldco%2Fpromptui%22&type=code

displague:
Closed via #407 

Along the way, #399 started an exploration into using more advanced UI toolkits including TUIs which could be part of an answer to #52.

PR #399, opened 2023-11-29
fix: deprecated promptui library

codinja1188:
# What fixes the PR?

https://github.com/equinix/metal-cli/issues/331


internal/organizations/delete.go
```diff
@@ -58,19 +59,18 @@ func (c *Client) Delete() *cobra.Command {
...
-				prompt := promptui.Prompt{
-					Label:     fmt.Sprintf("Are you sure you want to delete organization %s: ", organizationID),
-					IsConfirm: true,
-				}
 
-				_, err := prompt.Run()
-				if err != nil {
+			if !force {
+				fmt.Printf("Are you sure you want to delete device %s (y/N): ", organizationID)
+				userInput := prompt.Input(">", func(d prompt.Document) []prompt.Suggest {
```
displague: In this PR, most of the prompts are not using suggestions. If we choose not to use the suggestion feature, we are better off using a native [fmt.Scanln](https://pkg.go.dev/fmt#Scanln) and avoiding any dependency.

On the other hand, prompt suggestions ([shown here](https://github.com/c-bata/go-prompt#go-prompt) and implemented (inconsistently) for [port convert below](https://github.com/equinix/metal-cli/pull/399/files#diff-04080aab6a2c5bc5b599c51a90e52ef2ba3a6244b39a11cca412f92423bdd9bfR70)) could offer a better interface for our `metal init` steps and could be used when a user is present to prompt (with suggestions) when a required value has been omitted (we can detect a terminal or stdin).

displague: In other words, let's go all in and introduce this to the `metal init` prompts and use yes/no suggestions for all confirmation prompts, or lets just use fmt.Sscanln and worry about more featureful UIs later (I thought we had an issue for that, suggesting TUI toolkits, but I don't see one).

This c-bata/go-prompt package could work for our input needs, but we may want to opt for something more comprehensive for our output needs. We don't offer field selection options in our table responses and we don't offer watch views. We could move in that direction with a package like https://github.com/rivo/tview or by adopting whatever mechanisms kubectl is using to support it's set feature set of tsv, jq, go template, json, yaml, and other output methods.

displague: I'd want to avoid introducing a fancy library (picker, autocomplete, dropdown, etc) to replace it with an aesthetically different one soon after. Going from our current prompt library to fmt.Scanln would avoid that with no user-facing change in experience.

codinja1188: I will check the feasibility to adopt tview

codinja1188:
```
if !force {
		reader := bufio.NewReader(os.Stdin)
		fmt.Printf("Are you sure you want to delete device %s? (yes/no): ", deviceID)
		confirmation, err := reader.ReadString('\n')
		if err != nil {
			return fmt.Errorf("error reading confirmation: %w", err)
		}
		if strings.TrimSpace(strings.ToLower(confirmation)) != "yes" {
			fmt.Println("Device deletion cancelled.")
			return nil
		}
	}
```

codinja1188: @displague @ctreatma @cprivitere 

plz confirm the above approach

displague:
The initial concern of this PR has been addressed by #407 

We discussed treating this PR as an experiment towards the introduction of some future TUI. We don't have an issue open for that at present and we aren't ready to consider alternatives at the moment.

PR #405, opened 2023-12-11
fix: Add cleanups in device,vlan and ports e2e tests to stabilise the CI pipeline

### LINKED ISSUE #343: E2E tests (run in CI) are not cleaning up server and project artifacts
### What happened?

Servers created through E2E testing are not being deleted after the tests complete.

The project name is the same in each test so it is difficult to say which test triggered the project (and to identify if the devices were orphaned as a result of a failed run).

It is possible that an attempt was made to delete the servers and this failed (on the API side). This would block the deletion of the project.

### How can we reproduce it?

Ensure that deletion is always carried out on the servers in the project and the project.

Create test servers with a deletion value so that they are automatically reaped.

CLI version (`metal --version`):

### LINKED ISSUE #387: TestPorts_VLANs E2E test failed

aayushrangwala:
fixes https://github.com/equinix/metal-cli/issues/343
fixes https://github.com/equinix/metal-cli/issues/387

### DIFF
--- a/test/e2e/capacitytest/capacity_test.go
+++ b/test/e2e/capacitytest/capacity_test.go
@@ -41,12 +41,16 @@ func TestCli_Capacity(t *testing.T) {
 				rescueStdout := os.Stdout
 				r, w, _ := os.Pipe()
 				os.Stdout = w
+				t.Cleanup(func() {
+					w.Close()
+					os.Stdout = rescueStdout
+				})
+
 				if err := root.Execute(); err != nil {
-					t.Error(err)
+					t.Fatal(err)
 				}
-				w.Close()
+
 				out, _ := io.ReadAll(r)
-				os.Stdout = rescueStdout
 				if !strings.Contains(string(out[:]), "n3.xlarge.x86") &&
 					!strings.Contains(string(out[:]), "m3.large.x86") &&
 					!strings.Contains(string(out[:]), "s3.xlarge.x86") &&
@@ -70,12 +74,16 @@ func TestCli_Capacity(t *testing.T) {
 				rescueStdout := os.Stdout
 				r, w, _ := os.Pipe()
 				os.Stdout = w
+				t.Cleanup(func() {
+					w.Close()
+					os.Stdout = rescueStdout
+				})
+
 				if err := root.Execute(); err != nil {
-					t.Error(err)
+					t.Fatal(err)
 				}
-				w.Close()
+
 				out, _ := io.ReadAll(r)
-				os.Stdout = rescueStdout
 				if !strings.Contains(string(out[:]), "c3.small.x86") &&
 					!strings.Contains(string(out[:]), "mt") &&
 					!strings.Contains(string(out[:]), "sv") &&
@@ -99,12 +107,16 @@ func TestCli_Capacity(t *testing.T) {
 				rescueStdout := os.Stdout
 				r, w, _ := os.Pipe()
 				os.Stdout = w
+				t.Cleanup(func() {
+					w.Close()
+					os.Stdout = rescueStdout
+				})
+
 				if err := root.Execute(); err != nil {
-					t.Error(err)
+					t.Fatal(err)
 				}
-				w.Close()
+
 				out, _ := io.ReadAll(r)
-				os.Stdout = rescueStdout
 				if !strings.Contains(string(out[:]), "m3.large.x86") &&
 					!strings.Contains(string(out[:]), "mt") &&
 					!strings.Contains(string(out[:]), "sv") &&
@@ -127,13 +139,16 @@ func TestCli_Capacity(t *testing.T) {
 				rescueStdout := os.Stdout
 				r, w, _ := os.Pipe()
 				os.Stdout = w
+				t.Cleanup(func() {
+					w.Close()
+					os.Stdout = rescueStdout
+				})
+
 				if err := root.Execute(); err != nil {
-					t.Error(err)
-					return
+					t.Fatal(err)
 				}
-				w.Close()
+
 				out, _ := io.ReadAll(r)
-				os.Stdout = rescueStdout
 				if !strings.Contains(string(out[:]), "c3.medium.x86") &&
 					!strings.Contains(string(out[:]), "ny") &&
 					!strings.Contains(string(out[:]), "da") {
@@ -154,13 +169,16 @@ func TestCli_Capacity(t *testing.T) {
 				rescueStdout := os.Stdout
 				r, w, _ := os.Pipe()
 				os.Stdout = w
+				t.Cleanup(func() {
+					w.Close()
+					os.Stdout = rescueStdout
+				})
+
 				if err := root.Execute(); err != nil {
-					t.Error(err)
-					return
+					t.Fatal(err)
 				}
-				w.Close()
+
 				out, _ := io.ReadAll(r)
-				os.Stdout = rescueStdout
 				if !strings.Contains(string(out[:]), "c3.medium.x86") &&
 					!strings.Contains(string(out[:]), "m3.large.x86") &&
 					!strings.Contains(string(out[:]), "ny") &&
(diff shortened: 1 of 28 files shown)


test/e2e/capacitytest/capacity_test.go
```diff
@@ -129,6 +129,7 @@ func TestCli_Capacity(t *testing.T) {
 				os.Stdout = w
 				if err := root.Execute(); err != nil {
 					t.Error(err)
+					return
```
displague: I believe that all of the occurrences of:
```
t.Error(err)
return
```

can be replaced by the following which provides additional logging context:
```
t.Fatal(err)
```

https://pkg.go.dev/testing#T.Fatal

displague: A few more points are discussed here, https://chat.openai.com/share/4d97fa93-08b7-4e4d-8a7c-2343394a6ffe
- we could handle the Close() and remapping of Stdout in a defer call so that it happens in error conditions
- there are potential error conditions that are not being checked by the tests

test/helper/helper.go
```diff
@@ -433,3 +435,7 @@ func GenerateRandomString(length int) (string, error) {
 
 	return randomString, nil
 }
+
+func GenerateUUID() string {
```
ctreatma: Let's use the existing `GenerateRandomString` function (defined above) rather than adding a second function to serve the same purpose.

test/e2e/devices/devicecreatetest/device_create_test.go
```diff
@@ -42,46 +42,65 @@ func TestCli_Devices_Create(t *testing.T) {
...
+				})
 				if err != nil {
 					t.Error(err)
+					return
 				}
+
 				if len(projectId) != 0 {
 
-					root.SetArgs([]string{subCommand, "create", "-p", projectId, "-P", "m3.small.x86", "-m", "da", "-O", "ubuntu_20_04", "-H", "metal-cli-create-dev"})
+					deviceName := "metal-cli-create-dev" + helper.GenerateUUID()
```
ctreatma: The random string should be generated once at the start of a test, and the same value should be used in all resource names for the test (so the test project and the test devices in that project should have the same random string in their names).

test/helper/helper.go
```diff
@@ -312,21 +315,22 @@ func CleanupProjectAndDevice(deviceId, projectId string) error {
 //nolint:staticcheck
 func SetupProjectAndDevice(t *testing.T, projectId, deviceId *string) *openapiclient.Device {
 	t.Helper()
-	projId, err := CreateTestProject("metal-cli-test-project")
+	projectName := "metal-cli-test-project" + GenerateUUID()
```
ctreatma: The project name prefix here should be passed in, not hard-coded to `metal-cli-test-project`; I see a couple orphaned projects from these tests, but I can't tell which tests they came from because of the generic prefix.

test/e2e/capacitytest/capacity_test.go
aayushrangwala: Thanks @displague for sharing those. For using `t.Fatal()` it will abruptly stops the execution hence the `t.Cleanup()` or `defer` wont be executed
chatgpt response
```In Go's testing package, if t.Fatal is called during a test, it stops the execution of that test function immediately. This means that any subsequent code, including deferred functions like t.Cleanup, won't be executed.

The purpose of t.Fatal is to indicate that the test has failed and further test execution is unnecessary because the conditions for a successful test are not met. When t.Fatal is called, it ends the test function abruptly, preventing subsequent statements or deferred functions (including t.Cleanup) from being executed.```

Comment (displague) [reply test/e2e/capacitytest/capacity_test.go]: @aayushrangwala good call using t.Cleanup instead of defer. 👍

File: test/helper/helper.go (hunk 2 above)
Comment (aayushrangwala) [reply test/helper/helper.go]: Ah right didnt notice as it recently got merged I guess

Comment (aayushrangwala) [reply test/helper/helper.go]: Done

File: test/e2e/devices/devicecreatetest/device_create_test.go (hunk 3 above)
Comment (aayushrangwala) [reply test/e2e/devices/devicecreatetest/device_create_test.go]: Right. Fixed

File: test/helper/helper.go (hunk 4 above)
Comment (aayushrangwala) [reply test/helper/helper.go]: Yes right. Fixed

File: test/helper/helper.go [hunk 5]
```diff
@@ -405,14 +408,13 @@ func CreateTestBgpEnableTest(projId string) error {
...
-		fmt.Fprintf(os.Stderr, "Error when calling `BGPApi.RequestBgpConfig``: %v\n", err)
-		return err
+		return fmt.Errorf("Error when calling `BGPApi.RequestBgpConfig``: %v\n", err)
 	}
 	return nil
 }
 
 //nolint:staticcheck
-func GenerateRandomString(length int) (string, error) {
+func GenerateRandomString(length int) string {
```
ctreatma, line 420: The signature and implementation of this function shouldn't change; it should be used as-is.

test/helper/helper.go
```diff
@@ -431,5 +433,5 @@ func GenerateRandomString(length int) (string, error) {
 	// Trim any padding characters
 	randomString = randomString[:length]
```
displague: `panic: runtime error: slice bounds out of range [:5] with length 4`

The base64 encoding produces less characters as is unnecessary for this task. Here's a simplified approach that doesn't panic:

```go
func GenerateRandomString(length int) string {
	charSet := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
	rand.Seed(time.Now().UnixNano()) // seeding could happen outside of this function, in the test runner construction / init
	result := make([]byte, length)
	for i := range result {
		result[i] = charSet[rand.Intn(len(charSet))]
	}
	return string(result)
}
```

test/helper/helper.go
displague: The problem with the original function was that it would error out on smaller sized lengths - `5` for example. Somewhere in the history of this PR the random string length was modified from 32 to 5 and the E2E tests started failing.  We have seen more restrictive resource lengths introduced in the EM API, specifically around tags. I don't know if that was a motivation for the initial change.

Once the E2E tests started failing, I took a look at what would be needed to fix the base64-based GenerateRandomString function. The complexity hid the problem. As rewritten, the function can not fall into an error condition so the signature was modified to remove the unused error.  If we are ok with accepting that change in this PR, we'll see the echo of that change in this PR.

https://go.dev/play/p/Z2zkdK5lSEU

displague (approved):
approving contingent on tests (which seem to be stuck)

displague:
E2E test failures were about availability and the ability to access resources that were no longer available. Nothing tied to these changes.

displague:
I was anxious to capture the testing improvements and refactors and my last comment (and merge) was oblivious to the premise of this PR, to address the device-related E2E test failures.  We still have some and I'll reopen the tracking issue to say as much.

PR #421, opened 2023-12-20
test: use TerminationTime to sweep devices in E2E tests

displague:
The E2E tests have not been good about cleaning up their devices and projects. By taking advantage of the TerminationTime feature of Equinix Metal device deployments, we can ensure that the resources are swept up even when our sweeping processes fail.

1 Hour was chosen because this is longer than we permit for E2E tests in metal-cli today. (I've seen 15m timeouts)

A similar process is used for Terraform E2E tests.


codinja1188 (approved):
/LGTM

ctreatma (approved):

PR #428, opened 2024-01-05
chore: add User-Agent prefix to test helper client

ctreatma:
Adding a different User-Agent for the test helper client will help us correctly identify traffic that was generated from running CLI commands as opposed to traffic that was generated by using the SDK directly.


ctreatma:
This was broken out of #423 and should be merged before that PR.

displague (approved):

PR #431, opened 2024-01-10
refactor: use SDK-provided paginators

ctreatma:


internal/ips/retrieve.go
```diff
@@ -101,11 +100,12 @@ func (c *Client) Retrieve() *cobra.Command {
 				return c.Out.Output(ip, header, &data)
 			}
 
-			ips, err := pager.GetAllIPReservations(c.IPService, projectID, inc, exc, types)
+			// Don't have a paginator for this one
+			resp, _, err := c.IPService.FindIPReservations(context.Background(), projectID).Include(inc).Exclude(exc).Types(types).Execute()
```
ctreatma: Seems like we need to patch in a paginator for the IP reservations endpoint (and maybe others?).  IIRC we had a python script in metal-go for identifying paginatable endpoints so maybe we can set up equinix-sdk-go to automatically find and patch these things.

internal/events/retrieve.go
```diff
@@ -70,39 +68,39 @@ func (c *Client) Retrieve() *cobra.Command {
 				return fmt.Errorf("id, project-id, device-id, and organization-id parameters are mutually exclusive")
 			} else if deviceID != "" {
 				deviceRequest := c.EventService.FindDeviceEvents(context.Background(), deviceID).Include(c.Servicer.Includes(inc)).Exclude(c.Servicer.Excludes(nil))
-				events, err = pager.GetDeviceEvents(deviceRequest)
+				resp, err := deviceRequest.ExecuteWithPagination()
 				if err != nil {
 					return fmt.Errorf("could not list Device Events: %w", err)
 				}
+				events = resp.Events
```
ctreatma, line 75: This could be made less repetitive but I have chosen not to do that here because it would require more extensive changes to the existing if/else/etc. structure.

internal/ips/retrieve.go
ctreatma: Actually, on closer inspection the old CLI-specific paginator was also not paginating; if it ever got a response that had more than one page it would just duplicate the first page over and over: https://github.com/equinix/metal-cli/blob/671da79d4fab26d68b0f430cd0ad12e35bb738f6/internal/pagination/pager.go#L155

The page parameter isn't included in the spec for the find IP reservations endpoint but I don't know if that's a spec bug or if it's intentional; @sbhatnagar-equinix do you know if the IP reservations endpoint is supposed to support pagination?

ctreatma: Related issue in equinix-sdk-go: equinix/equinix-sdk-go#18

displague: We're not using the Servicer.Excludes/Includes functions in some of these pagination locations. This is needed to ensure that the `--include` and `--exclude` arguments from the CLI make it into the request.

This PR has me wondering if we should introduce `--page`, `--per-page` as globals, like `include` and `exclude`. These would also be exposed through the Servicer to pass the desired pagination settings into the request. This would allow a user to skip the full list or paginate on their own terms. Sort options (`sort_by`, `sort_dir`) would also make sense with `--page`.

ctreatma: I wired up the missing includes/excludes flags for the commands that are touched in this PR, but there may be others that were missed in the SDK conversion.

IMO once someone wants deep control of pagination they're probably better off using the SDK directly, rather than coding around the CLI, but we could open an issue for that feature and solicit interest/feedback from users.

ctreatma: Worked around this issue for now by upgrading to 0.31.2, but the upstream defect still exists.

displague (approved):

PR #432, opened 2024-01-11
fix: updated apiURL for e2e tests

codinja1188:
Updated apiURL for all tests


ctreatma:
Let's also include in this PR the change to `cli/root.go` to pass the API URL in to the `equinix-sdk-go` client.

codinja1188:
> Let's also include in this PR the change to `cli/root.go` to pass the API URL in to the `equinix-sdk-go` client.

Updated. 

@ctreatma can you approve the e2e workflow

ctreatma (approved):
I see one test failure and it looks like a provisioning issue (in any case it's an API error which means the test is successfully talking to the API).

Issue #414, opened 2023-12-15
Add support for Metal VRF Routes


displague:
### What problem are you facing?

With #268 there are a few features of VRF yet to be implemented. Among them is VRF Routes configuration.

https://github.com/equinix-labs/metal-go/blob/main/docs/VRFsApi.md

Method | HTTP request | Description
------------- | ------------- | -------------
[**CreateVrfRoute**](https://github.com/equinix-labs/metal-go/blob/main/docs/VRFsApi.md#CreateVrfRoute) | **Post** /vrfs/{id}/routes | Create a VRF route
[**DeleteVrfRouteById**](https://github.com/equinix-labs/metal-go/blob/main/docs/VRFsApi.md#DeleteVrfRouteById) | **Delete** /routes/{id} | Delete a VRF Route
[**FindVrfRouteById**](https://github.com/equinix-labs/metal-go/blob/main/docs/VRFsApi.md#FindVrfRouteById) | **Get** /routes/{id} | Retrieve a VRF Route
[**GetVrfRoutes**](https://github.com/equinix-labs/metal-go/blob/main/docs/VRFsApi.md#GetVrfRoutes) | **Get** /vrfs/{id}/routes | Retrieve all routes in the VRF
[**UpdateVrfRouteById**](https://github.com/equinix-labs/metal-go/blob/main/docs/VRFsApi.md#UpdateVrfRouteById) | **Put** /routes/{id} | Update a VRF Route

### How could the Equinix Metal CLI help solve this problem?

```
metal vrf routes-create --vrf-id {vrf_id}
metal vrf routes-update --route-id {route_id}
metal vrf routes-get --route-id {route_id}
metal vrf routes-delete --route-id {route_id}
```

One of these two ids should probably be "id" as an argument. Use whatever the current convention is for subcommands with IDs and when we use the short "id" vs the subcommand prefix.

cprivitere:
When we create a vrf currently, we get the following output:

```
❯ metal vrf create  -m ch -r 192.168.1.1/29 -a 12003 -n testvrf
+---------+------+-------------+----------------+-------------------------------+---------+
|   ID    | NAME | DESCRIPTION |    LOCALASN    |           IPRANGES            | CREATED |
+---------+------+-------------+----------------+-------------------------------+---------+
| testvrf |      | 12003       | 192.168.1.0/29 | 2024-01-19 20:33:52 +0000 UTC |
+---------+------+-------------+----------------+-------------------------------+---------+
```

[...]

PR #441, opened 2024-01-30
fix: handle optional flags correctly in VRF route update

ctreatma:
The test for the VRF route update subcommand was explicitly skipped.  I added a helper function to wait for the VRF route to be ready before testing the update subcommand, but when the test was enabled it failed because the subcommand was not capable of updating tags without updating all other VRF route properties.

In addition to enabling the test for the VRF route update subcommand, this updates the subcommand itself so that it only updates a VRF route property if the user set the corresponding CLI flag.

Part of #414

### DIFF
--- a/test/helper/helper.go
+++ b/test/helper/helper.go
@@ -197,6 +197,33 @@ func WaitForDeviceState(t *testing.T, deviceId string, states ...metalv1.DeviceS
 	return false, fmt.Errorf("timed out waiting for device %v state %v to become one of %v", deviceId, device.GetState(), states)
 }
 
+func WaitForVrfRouteState(t *testing.T, routeId string, statuses ...metalv1.VrfRouteStatus) bool {
+	var route *metalv1.VrfRoute
+	var err error
+	t.Helper()
+	predefinedTime := 900 * time.Second // Adjust this as needed
+	retryInterval := 10 * time.Second   // Adjust this as needed
+	startTime := time.Now()
+	client := TestClient()
+	for time.Since(startTime) < predefinedTime {
+		route, _, err = client.VRFsApi.FindVrfRouteById(context.Background(), routeId).Execute()
+		if err != nil {
+			t.Fatal(err)
+			return false
+		}
+		for _, status := range statuses {
+			if route.GetStatus() == status {
+				return true
+			}
+		}
+
+		// Sleep for the specified interval
+		time.Sleep(retryInterval)
+	}
+	t.Fatalf("timed out waiting for VRF route %v status %v to become one of %v", routeId, route.GetStatus(), statuses)
+	return false
+}
+
 func WaitForAttachVlanToPort(t *testing.T, portId string, attach bool) error {
 	t.Helper()
 	ticker := time.NewTicker(5 * time.Second)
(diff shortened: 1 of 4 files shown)


cprivitere:
Looks like the vrf-route-update-test is panicking after 30 seconds.

displague:
There have been conversations about adding a `--wait` argument for `create` subcommands. The code used in testing could be surfaced as a feature in the future.

ctreatma:
> Looks like the vrf-route-update-test is panicking after 30 seconds.

I see a panic happening in a different test due to the test timeout being reached (~1000 seconds) and the VRF route update test failing because the route isn't ready in time.  I triggered a re-run of the tests to see if the route update times out again.  We may need to extend the test timeout ([terraform uses `180m`](https://github.com/equinix/terraform-provider-equinix/blob/main/.github/workflows/acctest.yml#L112))

test/helper/helper.go
```diff
@@ -188,6 +188,33 @@ func WaitForDeviceState(t *testing.T, deviceId string, states ...metalv1.DeviceS
 	return false, fmt.Errorf("timed out waiting for device %v state %v to become one of %v", deviceId, device.GetState(), states)
 }
 
+func WaitForVrfRouteState(t *testing.T, routeId string, statuses ...metalv1.VrfRouteStatus) bool {
+	var route *metalv1.VrfRoute
+	var err error
+	t.Helper()
+	predefinedTime := 900 * time.Second // Adjust this as needed
```
ctreatma: I observed this test passing at least once locally, but in CI it's timed out twice in a row.  Can it really take more than 15 minutes for a VRF route to finish updating or are we doing something wrong here?

test/helper/helper.go
```diff
@@ -188,6 +188,33 @@ func WaitForDeviceState(t *testing.T, deviceId string, states ...metalv1.DeviceS
...
+	retryInterval := 10 * time.Second   // Adjust this as needed
+	startTime := time.Now()
+	client := TestClient()
+	for time.Since(startTime) < predefinedTime {
+		route, _, err = client.VRFsApi.FindVrfRouteById(context.Background(), routeId).Execute()
+		if err != nil {
+			t.Fatal(err)
+			return false
+		}
+		for _, status := range statuses {
```
displague: We could be using one of these here:
- https://pkg.go.dev/slices@go1.21.6#Contains 
- https://pkg.go.dev/slices@go1.21.6#ContainsFunc

(added in 1.18, metal-cli is at 1.19)

test/helper/helper.go
displague: What was the pending status during the failure?

test/helper/helper.go
ctreatma: The waiter that was added here is not needed (and in fact makes the test fail).  In terms of future cleanup, if we want to update similar code blocks in this repo to use slices, we either have to update to go1.21 so we get it from stdlib or we need to pull in golang.org/x/exp/slices.

test/helper/helper.go
ctreatma: Per offline conversation, the VRF route status will be pending until some other setup is done.  The pending status should not be used to determine if an update request can be made.  There is a hard-coded 5-minute wait before an update request is allowed, and the tests have been updated to match that.

It's also possible this entire VRF route subcommand should be removed.

cprivitere (approved):
Tests are flakey but passing in general. The ones added in this PR pass and that's what matters.

PR #442, opened 2024-01-31
fix: Updated metal-go client for sub-command facilities

codinja1188:


internal/facilities/retrieve.go
```diff
@@ -37,23 +24,56 @@ func (c *Client) Retrieve() *cobra.Command {
...
+				data[i] = []string{facility.GetName(), facility.GetCode(), metro, strings.Join(([]string)(stringFeatures), ",")}
 			}
 			header := []string{"Name", "Code", "Metro", "Features"}
-
 			return c.Out.Output(facilities, header, &data)
 		},
 	}
+
+	retrieveFacilitesCmd.Flags().StringVarP(&projectID, "project-id", "p", "", "The project's UUID. This flag is required, unless specified in the config created by metal init or set as METAL_PROJECT_ID environment variable.")
+	retrieveFacilitesCmd.Flags().StringVarP(&organizationID, "organization-id", "O", "", "UUID of the organization")
```
ctreatma: It looks like the packngo implementation was using the equivalent of [this endpoint](https://github.com/equinix/equinix-sdk-go/blob/main/services/metalv1/api_facilities.go#L59) and did not support CLI flags.  We should replicate what the packngo implementation did instead of adding new features & changing default behavior.

ctreatma (changes requested):

internal/facilities/retrieve.go
codinja1188: Do you mean only above mentioned endpoint support is enough?

codinja1188: By default our changes supports only above mentioned endpoint

internal/facilities/retrieve.go
```diff
@@ -37,23 +20,39 @@ func (c *Client) Retrieve() *cobra.Command {
...
+			)
 			cmd.SilenceUsage = true
-			facilities, _, err := c.Service.List(c.Servicer.ListOptions(nil, nil))
+
+			inc := []metalv1.FindFacilitiesIncludeParameterInner{}
+			exc := []metalv1.FindFacilitiesIncludeParameterInner{}
+			facilityList, _, err = c.Service.FindFacilities(context.Background()).Include(inc).Exclude(exc).Execute()
 			if err != nil {
-				return fmt.Errorf("Could not list Facilities: %w", err)
+				return fmt.Errorf("could not list All Metal Facilities: %w", err)
```
ctreatma: We don't need to change this error message

codinja1188: updated

ctreatma (approved):

PR #471, opened 2024-06-05
chore: convert to push-button release workflow

### LINKED ISSUE #364: Increase automation of release job
### What problem are you facing?

At the moment, the release process for `metal-cli` is fairly manual:

- Look at the `<previous>` release version
- Look at the commits to `main` since the previous release
- Decide on the `<next>` release version
- Create a tag on `main` in the format `v<next>`
- Push the tag to GitHub if it was created locally
- Tag creation triggers [a GitHub workflow that runs goreleaser](https://github.com/equinix/metal-cli/blob/eb835d0c0b9e86e3cc139522d57336bb7e2a79e2/.github/workflows/release.yml)

Now that we are ensuring that the commit log has [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) tags, we could move all of these steps into the release job so that releasing this CLI is just a matter of pushing a button and letting GitHub Actions handle all of the details.

ctreatma:
This updates the existing release workflow so that it is run by pushing a button in the GitHub UI rather than creating a tag locally and pushing that tag up to GitHub.

The push-button workflow looks up the previous release and uses Conventional Commit tags to determine the correct next version for the new release.  As before, it uses GoReleaser to build the CLI and upload the binaries to the GitHub release.

I have confirmed on my fork that this works up through the GoReleaser step.

Closes #364

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,26 +1,30 @@
-name: release
+name: Generate Next Release
 
 on:
-  push:
-    tags:
-    - '*'
+  workflow_dispatch:
 
 jobs:
-  goreleaser:
+  release:
     runs-on: ubuntu-latest
     steps:
-      -
-        name: Checkout
+      - name: Checkout
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      -
-        name: Set up Go
+      - name: Set up Go
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-      -
-        name: Run GoReleaser
+      - name: Release
+        id: semantic-release
+        uses: cycjimmy/semantic-release-action@v4
+        with:
+          semantic_version: 19.0.5
+          extra_plugins: |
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -1,30 +1,34 @@
...
-      -
-        name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v5
+      - name: Release
+        id: semantic-release
+        uses: cycjimmy/semantic-release-action@v4
         with:
-          version: latest
-          args: release --rm-dist
+          semantic_version: 19.0.5
```
ocobles: what's the reason to keep 19.0.5? https://github.com/semantic-release/semantic-release/releases/tag/v24.0.0

ctreatma: There's no reason any more.  I bumped this version and the conventional commits changelog plugin version.

displague (approved):

PR #476, opened 2024-06-25
feat: add IP addresses to the metal device get output

### LINKED ISSUE #242: Give IP address in server output table
When a user asks for "metal devices list" the IP address is not provided in the table which makes it hard for the user to access the device.

ajmaidak:
Closes: https://github.com/equinix/metal-cli/issues/242

### DIFF
--- a/internal/devices/retrieve.go
+++ b/internal/devices/retrieve.go
@@ -24,7 +24,6 @@ import (
 	"context"
 	"fmt"
 	"strconv"
-	"strings"
 
 	"github.com/equinix/equinix-sdk-go/services/metalv1"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


internal/devices/retrieve.go
```diff
@@ -52,56 +54,59 @@ func (c *Client) Retrieve() *cobra.Command {
...
-
 			for i, dc := range devices {
-				data[i] = []string{dc.GetId(), dc.GetHostname(), dc.OperatingSystem.GetName(), fmt.Sprintf("%v", dc.GetState()), dc.GetCreatedAt().String()}
+				ips := make([]string, 0)
+				for _, ip := range dc.GetIpAddresses() {
+					if ip.Address != nil {
+						ips = append(ips, *ip.Address)
+					}
+				}
+				data[i] = []string{dc.GetId(), dc.GetHostname(), strings.Join(ips, ","), dc.OperatingSystem.GetName(), fmt.Sprintf("%v", dc.GetState()), dc.GetCreatedAt().String()}
```
ctreatma: The original issue wasn't really clear on which IP(s) to include in the output; the [Metal console](https://console.equinix.com/) shows only the Public IPv4 address, so it's possible the goal was to show that address to bridge the gap between this and the web view.

That said, I think it's useful to include all of the IPs, but there's some important information lost by putting them in a single field.  Whether we show a single IP or show multiple, I think we should have a separate column per IP type with a header that identifies the IP address type.  The IP address types can be determined by looking at `ip.GetAddressFamily()` (4 for IPv4, 6 for IPv6) and `ip.GetPublic()`(true if it's a public address, false if private).

ajmaidak: I added columns for the public/private ipv4 and ipv6 addresses.

ctreatma: Nice!

ctreatma (approved):

Issue #106, opened 2021-02-20
Generate the CLI tool


displague:
With github.com/t0mk/gometal as a guide, it should be possible to generate a client from the spec:

https://github.com/go-swagger/go-swagger/issues/139#issuecomment-624340719
* https://github.com/jsilland/sutro
* https://github.com/jsilland/go-swagger-cli

displague:
Other projects with OpenAPI generated CLIs:
- [Restish](https://rest.sh/#/guide?id=api-specific-commands)
- [Linode CLI](https://github.com/linode/linode-cli/blob/fedb0e83c9b63a8d9f36e2b98e6a511c1ba52d1e/linodecli/operation.py#L201-L207)
- https://github.com/danielgtaylor/openapi-cli-generator

displague:
https://nordicapis.com/auto-generating-a-cli-from-openapi-specification/ discusses https://github.com/danielgtaylor/openapi-cli-generator.

https://github.com/danielgtaylor/openapi-cli-generator reports that it has been replaced by https://rest.sh/ (https://github.com/danielgtaylor/restish)

PR #485, opened 2024-08-26
fix: "device update --locked" should also unlock

### LINKED ISSUE #484: bug: command `--locked false` to unlock the server doesn't work, however `--locked true` to lock the server does work.
### What happened?

The `--locked false` and `-l false` commands do not work. According to [the Docs](https://deploy.equinix.com/developers/docs/metal/libraries/cli/#metal-device-update), a boolean can lock or unlock the server from being updated.

However, only `--locked true` and `-l true` seem to work. This is seen in the [video below](#video).

### How can we reproduce it?
1) Deploy server.
2) Get server ID.
3) Run command `metal device update --id <server-id> --locked true` to lock the server.
4) Confirm that it is locked on [console.equinix.com](console.equinix.com).

image-locked

5) Run command `metal device update --id <server-id> --locked false` to unlock the server.
6) Confirm that it is unlocked on [console.equinix.com](console.equinix.com). However, it is not.

image-unlocked

<br/>

> [!NOTE]
CLI ==version== (`metal --version`): `metal version 0.24.0`

<a name="video"></a>
> [!IMPORTANT]
https://github.com/user-attachments/assets/7b8e64bd-e9cd-459c-805e-81bb0b040f63

displague:
Fixes #484 

Previous behavior was to set `{"locked":true}` in `PUT` requests whenever `--locked` was specified. This is because the presence of the flag itself is treated as a toggle. The value argument was ignored unless given as `--locked=true` or `--locked=false`. When `--locked=false` was given, the `PUT` request was `{}`, since the code required true values to add the field to the update request.

This change technically breaks the current ability to do `metal device update --locked --some-other-arg`. 

```console
$ PACKNGO_DEBUG=1 go run ./cmd/metal device update --locked false --locked true --id 1234
Error: parameter locked may only be set once
```

```console
$ PACKNGO_DEBUG=1 go run ./cmd/metal device update --locked --id 1234                    
Error: invalid argument "--id" for "-l, --locked" flag: strconv.ParseBool: parsing "--id": invalid syntax
...
```

```console
$ PACKNGO_DEBUG=1 go run ./cmd/metal device update --locked true  --id 1234              
PUT /metal/v1//devices/1234 HTTP/1.1
...
{"locked":true}
```

```console
$ PACKNGO_DEBUG=1 go run ./cmd/metal device update --locked false  --id 1234
PUT /metal/v1//devices/1234 HTTP/1.1
...
{"locked":false}
```

### DIFF
--- a/internal/devices/update.go
+++ b/internal/devices/update.go
@@ -45,7 +45,7 @@ func (c *Client) Update() *cobra.Command {
 	)
 	// updateDeviceCmd represents the updateDevice command
 	updateDeviceCmd := &cobra.Command{
-		Use:   `update -i <device_id> [-H <hostname>] [-d <description>] [--locked <boolean>] [-t <tags>] [-u <userdata> | --userdata-file <filepath>] [-c <customdata>] [-s <ipxe_script_url>] [--always-pxe=<true|false>]`,
+		Use:   `update -i <device_id> [-H <hostname>] [-d <description>] [--locked=<true|false>] [-t <tags>] [-u <userdata> | --userdata-file <filepath>] [-c <customdata>] [-s <ipxe_script_url>] [--always-pxe=<true|false>]`,
 		Short: "Updates a device.",
 		Long:  "Updates the hostname of a device. Updates or adds a description, tags, userdata, custom data, and iPXE settings for an already provisioned device. Can also lock or unlock future changes to the device.",
 		Example: `  # Updates the hostname of a device:
--- a/internal/devices/update.go
+++ b/internal/devices/update.go
@@ -76,14 +76,11 @@ func (c *Client) Update() *cobra.Command {
 			}
 
 			if cmd.Flag("locked").Changed {
-				locked, err := cmd.Flags().GetBoolSlice("locked")
+				locked, err := cmd.Flags().GetBool("locked")
 				if err != nil {
 					return fmt.Errorf("could not parse locked value: %w", err)
 				}
-				if len(locked) > 1 {
-					return fmt.Errorf("parameter locked may only be set once")
-				}
-				deviceUpdate.Locked = &locked[0]
+				deviceUpdate.Locked = &locked
 			}
 
 			if len(tags) > 0 {
@@ -125,7 +122,7 @@ func (c *Client) Update() *cobra.Command {
 	updateDeviceCmd.Flags().StringVarP(&description, "description", "d", "", "Adds or updates the description for the device.")
 	updateDeviceCmd.Flags().StringVarP(&userdata, "userdata", "u", "", "Adds or updates the userdata for the device.")
 	updateDeviceCmd.Flags().StringVarP(&userdataFile, "userdata-file", "", "", "Path to a userdata file for device initialization. Can not be used with --userdata.")
-	updateDeviceCmd.Flags().BoolSliceP("locked", "l", []bool{}, "Locks or unlocks the device for future changes (<true|false>).")
+	updateDeviceCmd.Flags().BoolP("locked", "l", false, "Locks or unlocks the device for future changes (<true|false>).")
 	updateDeviceCmd.Flags().StringSliceVarP(&tags, "tags", "t", []string{}, `Adds or updates the tags for the device --tags="tag1,tag2".`)
 	updateDeviceCmd.Flags().BoolVarP(&alwaysPXE, "always-pxe", "a", false, "Updates the always_pxe toggle for the device (<true|false>).")
 	updateDeviceCmd.Flags().StringVarP(&ipxescripturl, "ipxe-script-url", "s", "", "Add or update the URL of the iPXE script.")
--- a/internal/devices/update.go
+++ b/internal/devices/update.go
@@ -34,7 +34,6 @@ import (
 func (c *Client) Update() *cobra.Command {
 	var (
 		description   string
-		locked        bool
 		userdata      string
 		userdataFile  string
 		hostname      string
@@ -80,8 +79,15 @@ func (c *Client) Update() *cobra.Command {
 				deviceUpdate.Userdata = &userdata
 			}
 
-			if locked {
-				deviceUpdate.Locked = &locked
+			if cmd.Flag("locked").Changed {
+				locked, err := cmd.Flags().GetBoolSlice("locked")
+				if err != nil {
+					return fmt.Errorf("could not parse locked value: %w", err)
+				}
+				if len(locked) > 1 {
+					return fmt.Errorf("parameter locked may only be set once")
+				}
+				deviceUpdate.Locked = &locked[0]
 			}
 
 			if len(tags) > 0 {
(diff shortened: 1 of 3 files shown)


internal/devices/update.go
```diff
@@ -123,12 +129,11 @@ func (c *Client) Update() *cobra.Command {
 	updateDeviceCmd.Flags().StringVarP(&description, "description", "d", "", "Adds or updates the description for the device.")
 	updateDeviceCmd.Flags().StringVarP(&userdata, "userdata", "u", "", "Adds or updates the userdata for the device.")
 	updateDeviceCmd.Flags().StringVarP(&userdataFile, "userdata-file", "", "", "Path to a userdata file for device initialization. Can not be used with --userdata.")
-	updateDeviceCmd.Flags().BoolVarP(&locked, "locked", "l", false, "Locks or unlocks the device for future changes (<true|false>).")
+	updateDeviceCmd.Flags().BoolSliceP("locked", "l", []bool{}, "Locks or unlocks the device for future changes (<true|false>).")
```
displague: I initially tried `BoolP`, but even when I sent `--locked false`, `GetBool` would return `true`. I think it was treating `--locked` like an argument that takes no parameters, in which case `false` would have been an `args[]` argument. Sending additional `foo` args on the command line doesn't trigger any errors, which backs this theory.

displague: Looks like we would need the following to generate an error in that condition:
```
updateDeviceCmd.Args = cobra.NoArgs
```

internal/devices/update.go
```diff
@@ -123,12 +125,13 @@ func (c *Client) Update() *cobra.Command {
...
 	updateDeviceCmd.Flags().StringSliceVarP(&tags, "tags", "t", []string{}, `Adds or updates the tags for the device --tags="tag1,tag2".`)
 	updateDeviceCmd.Flags().BoolVarP(&alwaysPXE, "always-pxe", "a", false, "Updates the always_pxe toggle for the device (<true|false>).")
 	updateDeviceCmd.Flags().StringVarP(&ipxescripturl, "ipxe-script-url", "s", "", "Add or update the URL of the iPXE script.")
 	updateDeviceCmd.Flags().StringVarP(&customdata, "customdata", "c", "", "Adds or updates custom data to be included with your device's metadata.")
 	_ = updateDeviceCmd.MarkFlagRequired("id")
-
+	updateDeviceCmd.MarkFlagsMutuallyExclusive("userdata", "userdata-file")
```
displague, line 131:
```console
$ PACKNGO_DEBUG=1 go run ./cmd/metal device update --locked false  --id 1234 --userdata foo --userdata-file=/tmp/foo
Error: if any flags in the group [userdata userdata-file] are set none of the others can be; [userdata userdata-file] were all set
Usage:
...
```

Surprisingly, this does not affect `make generate-docs`.

internal/devices/update.go
```diff
@@ -123,12 +125,13 @@ func (c *Client) Update() *cobra.Command {
...
+	updateDeviceCmd.Args = cobra.NoArgs
```
displague, line 132:
```console
$ PACKNGO_DEBUG=1 go run ./cmd/metal device update --locked false  --id 1234 fkk fll                                
Error: unknown command "fkk" for "metal device update"
```

internal/devices/update.go
displague: I'll add that in this PR for this action.

I can see us wanting to one-shot that change across the repo since we don't accept args for most (if any) commands.

ctreatma: It seems cobra requires `=` to set a value for boolean params: https://github.com/spf13/cobra/issues/1606

In that case the syntax would be `--locked=false`; we would still have to update the command logic to correctly handle unlocking (it currently only does something if `locked` is true).  An alternative would be a second `--unlocked` boolean flag that is mutually exclusive with `locked` (my personal preference and seems to be the recommended style for cobra commands).  We've also talked about switching to a string param, although that would require specifying the value `--locked <value>` instead of using the shorthand `--locked`.

ctreatma:
Re: the `fix!` tag, we're at v0 so any change can be breaking regardless of which version number component was bumped.  The release action assumes that you don't want to stay v0, so the `fix!` tag will make the next release of this CLI a v1 release.  Is that the intent?

internal/devices/update.go
displague: I could also go that way. The documentation previously said that the field was boolean and that `--locked <true|false>` was the way to toggle it. This PR upholds the documentation.

In practice, `--locked=<true|false>`, `--locked <true|false>`, and `--locked` all had the same effect previously.
I could see taking the alternative approach, introducing a new `--unlocked`.

ctreatma: The usage is hard-coded in the `Use` property of the subcommand and is not automatically generated based on the calls to `updateDeviceCmd.Flags()`, so the docs don't necessarily accurately reflect usage.  If you look at the usage for the `--always-pxe` flag, that matches the documented behavior/expectations of cobra boolean flags and establishes the expectation that a boolean flag value must be set with `=`.  Given that there's already a flag in this subcommand that takes `--<flag>=true|false` there's precedent for aligning `--locked` with that rather than introducing a separate flag.

displague (commented):
Updated the behavior to expect `--locked` or `--locked=<true|false>` and revised the docs to match.

Using `--locked true|false` will result in an error:
```
Error: unknown command "false" for "metal device update"
```

internal/devices/update.go
```diff
@@ -80,7 +75,11 @@ func (c *Client) Update() *cobra.Command {
 				deviceUpdate.Userdata = &userdata
 			}
 
-			if locked {
+			if cmd.Flag("locked").Changed {
```
ctreatma, line 78: A test for lock/unlock would build confidence that we've fixed the issue

stevemar (commented):
IIRC, using true/false as allowed values for a flag is generally not typical convention for CLIs. There should be an opposite flag, like `--unlock` to flip it. (See https://stackoverflow.com/questions/52403065/argparse-optional-boolean)

internal/devices/update.go
displague: I started down a path of introducing a mock because I don't like the idea of spinning up and settling electrified and cooled rust to verify that a few bytes of command line argument trigger a few bytes of json in an HTTP call.

https://github.com/equinix/metal-cli/commit/221430d331c20b0232e18d13f2f3b73dbda0b9d4

The test in the sha is not clean or complete, but it does trigger the lock in the CLI args and look for the request to contain the expected JSON.

Alternatively, I could add steps to the existing "update" E2E to see if the resource `locked` state can be toggled on and off again (which would be needed for a successful cleanup).

Interested in your thoughts on a preferred path, @ctreatma.

displague:
@stevemar 
This PR addresses the problem with `--locked false` (space separated). This problem appears elsewhere in the project, for example, `--force false`. That will be corrected with #486. 

The precedent in the `metal-cli` is for bool's to have `--name=<true|false>`. 
There's at least one exception that I see:
* `--app` vs `--sms` for TFA

We may find other places where `--name=false` is used, where false is the default but must be explicitly sent in API calls. It's confused to specify and with Cobra, it is easy to miss during code review.

Metal CLI can certainly introduce more obvious patterns in future versions, such as named pairs (`lock`, `unlock`) or `--no-` prefixes.  It would be great to get a better bool toggle behavior from the upstream project(s) too. 

Related issues:
* https://github.com/spf13/cobra/issues/958
* https://github.com/spf13/pflag/issues/214

Another alternative that fits this particular parameter would be to introduce `lock` and `unlock` `metal device` actions: `metal device lock --id 1234` vs `metal device update --lock --id 1234`.

`metal port convert` is a complicated example of this. It introduces a separate action, rather than `metal port update --layer-2`, but it also uses a parameter to trigger a secondary action: `metal port convert --bond` 🙃. 

The complications owe to RPC-like port endpoints in the otherwise REST-y API. In REST API's the verb is the HTTP Action (`GET`, `PUT`, `DELETE` ...), it should not be in the path name (`/convert`, `/assign`, `/disbond`). 

Metal CLI could be more predictable about how it approaches API to CLI translation, if not dogmatic. Documentation in `CONTRIBUTING.md` or #106 could help with that.

--- 

_I created https://github.com/equinix/metal-cli/issues/487 from this feedback and to continue this conversation._

internal/devices/update.go
ctreatma: I don't have a strong preference either way for whether the test hits a live API or a mock.  If we're going to mock I think it's better to set up a mock API than to inject a mock HTTP transport, since that more closely matches real usage, reduces the risk of accidentally mocking out some custom behavior, and provides a more straightforward path to maybe eventually adopting a mock that is generated from the API spec.

displague: @ctreatma added in https://github.com/equinix/metal-cli/pull/485/commits/c6562a20d37b122be18c4c5c7f4a10ff1d9310e8

test/e2e/devices/deviceupdatetest/device_update_test.go
```diff
@@ -42,14 +42,31 @@ func TestCli_Devices_Update(t *testing.T) {
...
 						t.Error("expected output should include metal-cli-update-dev-test in the out string ")
 					}
 				}
+
+				root.SetArgs([]string{subCommand, "delete", "-i", device.GetId()})
+
+				out := helper.ExecuteAndCaptureOutput(t, root)
+
+				if !strings.Contains(string(out[:]), "not delete") {
+					t.Error("expected output should include 'not delete' in the out string due to device locking")
```
ctreatma, line 59: This test is failing here; haven't checked this PR out yet to get a better idea of which part of this is failing (Device is not locked?  Device is locked but the error message doesn't match the check on line 58?  Something else?).

https://github.com/equinix/metal-cli/actions/runs/10741510710/job/29792085636?pr=485#step:5:77

ctreatma: The delete subcommand waits for confirmation before deleting: https://github.com/equinix/metal-cli/blob/main/internal/devices/delete.go#L40-L41

This test will need to delete with the `-f` flag to bypass the confirmation prompt.

PR #506, opened 2025-08-28
feat: Deprecate flags for creating spot market instances

adammohammed:
This just marks the spot-market related flags as deprecated as the API no longer accepts requests for spot instances.

More info can be found here: https://docs.equinix.com/metal/#retiring-spot-market


stevemar (approved):
lgtm

